PR #14, opened 2022-11-03
Fix: Simplify the code to resize a node

li-xin-yi:
Avoid passing `setNodes` and `onRize` functions when rendering nodes, which helps fix bugs when sync the nodes' info between users.

### DIFF
--- a/ui/src/components/repo/graph.js
+++ b/ui/src/components/repo/graph.js
@@ -15,6 +15,7 @@ import ReactFlow, {
   MiniMap,
   Controls,
   Handle,
+  useReactFlow
 } from "react-flow-renderer";
 import Box from "@mui/material/Box";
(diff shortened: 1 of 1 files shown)


ui/src/components/repo/graph.js
```diff
@@ -48,6 +49,25 @@ const ScopeNode = memo(({ data, id, isConnectable, selected }) => {
   const [frame] = React.useState({
     translate: [0, 0],
   });
+  const { setNodes } = useReactFlow();
```
lihebi, line 52: I didn't know this; this is convenient.

lihebi (approved):
LGTM. Thanks Xinyi!

This also seems to fix a performance issue for resizing nested nodes.

PR #36, opened 2022-11-08
Feature: collaborative editing

li-xin-yi:
### On code editing:

1. Prefer pulling contents from the collab server than database. If no data shared on the server, it initializes code from the database.
2. See the modification on the code editor in real-time
3. Show a cursor with unique color for different users when they are typing, move your mouse to the cursor can see the name of the editor.

image

#### On node creation and deletion

View the node appearing or disappearing immediately on other users' side.

### On node dragging

Show the moving track and shadow the selected node in the corresponding user color

image

### On node selection and resizing

Each user maintain an independent state of node selection. You can also be synced the resizing operations from other users immediately.

### Data store

It only writes to the database when creating or deleting a pod. All truth about nodes are stored at the server to sync among users. Of course, you can save them by explicit button and auto saving.


lihebi (approved):
LGTM, thanks for adding this feature!

PR #59, opened 2022-11-18
pod ui improve

nibilin33:


ui/src/index.css
```diff
@@ -10,4 +10,11 @@ body {
 code {
   font-family: source-code-pro, Menlo, Monaco, Consolas, 'Courier New',
     monospace;
+}
+.react-monaco-editor-container {
+  box-sizing: border-box;
+  padding: 10px 0;
+}
+.monaco .monaco-editor-background, .margin{
```
Himself65, line 18: https://stackoverflow.com/questions/44766624/custom-background-color-in-monaco-editor

lihebi (commented):
Thanks!

I prefer not to use separate css files and classNames, because
1. classNames are typically defined once and used once in React components , no need to give it a name
2. I want the css to be close to the React component that uses it. Cross reference is painful.
3. CSS cannot have variables, but MUI’s sx property can.
So, let’s use sx property of MUI component https://mui.com/system/getting-started/the-sx-prop/

Also, I see no need to use other styling libraries (e.g., styled components) other than MUI. The sx property of MUI seems to be implemented based on styled component already.

Issue #51, opened 2022-11-14
Improve Performance


lihebi:
Possible overhead:
- unnecessary re-renders
- Monaco editor
- React flow
- Yjs sync

Issue #55, opened 2022-11-18
Pod UI improvements


forrestbao:
This is how the pod UI looks like now. 
![image](https://user-images.githubusercontent.com/438579/202609763-550b74f0-1494-4fcd-9598-c4a4ac389bd9.png)

**Suggested changes** 
1. [ ] Move the top/tool bar of the pod to the top-right corner -- slightly extruding to the top, like Google CoLab's design (screenshot below). When a cell is not focused or hovered over, it should be hidden. When a cell is, display it. Just as in Google CoLab. 
    
    ![image](https://user-images.githubusercontent.com/438579/202611486-94c37d91-f792-454c-92bd-5f269d4efa2d.png)
2. [x] In the tool bar, for every icon, when hovered over, display the tooltop on top of it rather than to the right of it, which may block users from seeing or selecting the next icon. 
2. [x] The red dot is ugly. Change it to a trash can icon. 
3. [x] Align the output behavior the same as Jupyter. Do not distinguish `stdout` and `Result: [0]:`. 
4. [x] Border for a pod: 
    a. [ ] Round corner 
    b. [ ] black line to gray line. 
    c. [ ] Show shadow around the pod being edited. Currently we only shadow a pod being moved on the canvas. 
5. [x] Number the pod's order in execution. Just like in Jupyter, e.g., `In [21]: ` and `Out [21]:`. 
6. [x] If a pod was executed without error last time, show a green checker, as in DeepNote. 

Please take a look at the designs of Google Colab ([this notebook by me](https://colab.research.google.com/drive/15repLULKxghG5FYIeBbxoL-47JyGoRXI#scrollTo=Bu8IhewrNwcr)) and Deepnote ([this notebook by me](https://deepnote.com/workspace/forrest-bao-8d3a-8e96c0f9-789e-4b51-b08f-8c67f93fd301/project/NLPclass-5185c4e4-930a-4349-b76b-be1834007069/notebook/LDA-d3cd48df628e43349658499d66eb8093)).

forrestbao:
@nibilin33 Has the auto hide of toolbars also implemented?

forrestbao:
[...]

PR #61, opened 2022-11-18
Fix: Unnecessary Re-rendering

li-xin-yi:
Change
1. Use lazy eval function instead of accessing pods directly. So the components using pod/pods will not re-render when any value changes in pods.
2. Memorize `MyMonaco` component with only `id` and `gitvalue` property. Prevent force rendering from parent components.
3. Clean up auto layout function for `codeNode`.

The following work later if available:
1. `ResultBlock` will be re-rendered when its parent `codeNode` was dragged. I think we can also memo it and set the execution number like out[num] in (https://github.com/codepod-io/codepod/issues/55) as a prop to force re-rendering it only when the result changes. I just remained it as origin, please consider combining this issue when implementing the execution number.

close https://github.com/codepod-io/codepod/issues/51


lihebi (approved):
Thanks!

PR #62, opened 2022-11-19
core: pod update

nibilin33:
Move the toolbar horizontally
add delete pod


ui/src/components/Canvas.tsx
```diff
@@ -378,27 +415,74 @@ const CodeNode = memo<Props>(({ data, id, isConnectable }) => {
...
               zIndex: 100,
+              padding: "0 10px",
             }}
           >
             <ResultBlock pod={pod} id={id} />
           </Box>
         )}
       </Box>
+      {/* moveable will effect editor */}
+      {/* {false && (
```
lihebi, line 449: This was removed in #61, so there appear to be some problems with the merge.

ui/src/components/canvas.module.css
```diff
@@ -0,0 +1,27 @@
+.result-status__success{
```
lihebi, line 1: As I commented [here](https://github.com/codepod-io/codepod/pull/59#pullrequestreview-1187057820), let's inline the styles.

lihebi (commented):
Thanks, Bilin! A few comments:

1. there appear to be some problems with merging
2. styles should be inline as I mentioned [here](https://github.com/codepod-io/codepod/pull/59#pullrequestreview-1187057820)
3. Please use [prettier](https://marketplace.visualstudio.com/items?itemName=esbenp.prettier-vscode) to format your code.

li-xin-yi:
Thanks for your awesome work. Many bugs in my code are fixed in your PR, thanks a lot. However, I fetched this branch and tested, some tiny issues show up on my side but I'm not sure what cause them. Can you help me investigate them?

1.  The size of `codeNode` fluctuates when editing on its editor.
2.  Even when the cursor is in the editor, the `codeNode` is selected and draggable by cursor movement. But we are not supposed to enable the drag operator when editing, we intend to only drag the pod by the drag handler we set.
3. the input index is not correct, it doesn't update after run. Refresh the page, they appear but are even out of order.

nibilin33:
@li-xin-yi ok, i will check this

forrestbao:
@lihebi @li-xin-yi 
I asked @nibilin33 to remove the shadow. It is buggy and is now redundant given now editor borders can be highlighted when focused or in editing. But if you have a reason to use shadow, please let me know.

li-xin-yi:
> @lihebi @li-xin-yi I asked @nibilin33 to remove the shadow. It is buggy and is now redundant given now editor borders can be highlighted when focused or in editing. But if you have a reason to use shadow, please let me know.

@forrestbao  Thanks for the advices. I meant to use the box shadow and its color to show which user is dragging or selecting the pod, which are of the same color with the remote cursor. This is also the awareness feature in collaborative editing. I know it may look awkward, but I have no idea how to show the awareness info for dragging/selecting operation of pod in a proper form. Can you give me any suggestion?

forrestbao:
> @forrestbao Thanks for the advices. I meant to use the box shadow and its color to show which user is dragging or selecting the pod, which are of the same color with the remote cursor. This is also the **awareness** feature in **collaborative editing**. I know it may look awkward, but I have no idea how to show the awareness info for dragging/selecting operation upon a pod in a proper form. Can you give me any suggestion?
> 
> BTW, in my experimental demo (https://github.com/li-xin-yi/sharing-edited-monaco-flow), I convey the awareness and movement information by setting the pod half-transparent and changing its background color. But it seems not so suitable in our project.

@li-xin-yi  Ah, i didn't notice the shadowing is for dragging indication. Maybe we can use border highlighting, and potentially bolding, for both code editing and dragging?

nibilin33:
@li-xin-yi the issues you mentioned fixed

ui/src/components/Canvas.tsx
```diff
@@ -200,15 +241,19 @@ function ResultBlock({ pod, id }) {
...
           ) : (
-            pod.result.text && (
-              <Box>
-                <Box sx={{ display: "flex" }} bgcolor="lightgray">
-                  Result: [{pod.result.count}]:
-                </Box>
-                <Box>
-                  <Box component="pre" whiteSpace="pre-wrap">
-                    {pod.result.text}
-                  </Box>
```
lihebi, line 211: We need to show pod.result.text. In general, when a code block executes in Jupyter, the value of the last expression should be printed. like this:

Screenshot 2022-11-20 at 4 25 36 AM

ui/src/components/Canvas.tsx
```diff
@@ -248,158 +289,186 @@ function ResultBlock({ pod, id }) {
...
+      });
+    }
+  }, []);
+  const onLayout = useCallback(({ height }) => {
+    const node = nodesMap.get(id);
+    if (node) {
+      node.style = { ...node.style, height };
+      nodesMap.set(id, node);
+    }
+  }, []);
```
lihebi, line 330: These were removed in #61. So there are still merging problems. There might be other places, it's tedious to check. If it is not easy to merge, maybe let's create a clean PR from the current `main` branch?

ui/package.json
```diff
@@ -32,6 +32,7 @@
     "react-icons": "^4.6.0",
     "react-monaco-editor": "^0.50.1",
     "react-moveable": "^0.40.0",
+    "react-resizable": "^3.0.4",
```
lihebi, line 35: Need this:

```
yarn add -D @types/react-resizable
```

Otherwise the compiler complains:

```
ERROR in src/components/Canvas.tsx:37:30
TS2307: Cannot find module 'react-resizable' or its corresponding type declarations.
    35 |
    36 | import Moveable from "react-moveable";
  > 37 | import { ResizableBox } from "react-resizable";
       |                              ^^^^^^^^^^^^^^^^^
    38 | import Ansi from "ansi-to-react";
    39 |
    40 | import { customAlphabet } from "nanoid";
```

ui/src/components/Canvas.tsx
lihebi: cc @forrestbao

lihebi:
The merging problems seem complex. There is a lot of code that was removed in #61 that got added back in this PR. It's pretty hard to catch what they are. Also, this PR is getting too large to review. This might create additional conflicts with other changes that Xinyi and I are working on.

Maybe let's split this PR into smaller pieces?

ui/package.json
forrestbao: I was wondering why...I shelled into the UI instance to manually install react-resizable to fix.

ui/src/components/Canvas.tsx
forrestbao: I was wondering on this too. At first I thought this was due to that CodePod does not return the evaluation of the last expression.

ui/src/components/Canvas.tsx
nibilin33: this is for width adjustment, it's different from removed one

Issue #58, opened 2022-11-18
Focused pod not executed 


forrestbao:
**What I saw** 

After changing focus to a new cell (`print_TA`) and pressing Shift+Enter, the previously focused cell (`import random`) is executed unexpectedly. I had to manually click the Green play button of the current cell to execute it. 

**What I should have seen**

1. A focused cell should be executed when Shift+Enter is pressed. 
2. A cell NOT focused should NOT be executed when Shift+Enter is pressed.  

**Multimedia evidence** 

See the video below. 

https://user-images.githubusercontent.com/438579/202622426-f1d0f106-b2e4-4332-879f-cead32bcdb09.mp4

nibilin33:
it's MonacoEditor's bug https://github.com/microsoft/monaco-editor/issues/2947
it always triggered the last created instance =_=

PR #70, opened 2022-11-21
fix: pod excution

nibilin33:
issue: [58](https://github.com/codepod-io/codepod/issues/58)
reason:   MonacoEditor's bug , it always triggered the lastest created instance
solution:  add global store to memorize currentEditor


lihebi (approved):
Thanks!

Issue #46, opened 2022-11-12
Project sharing


lihebi:
Now we have real time collaboration infra. We need to add features to allow users to share a repo.
- A "share" button (in sidebar or navbar) to share with other users by email.
- Proper read/write access control in api/src/resolvers, and
- Make sure real-time collaboration works with sharing.

lihebi:
A project can also be marked as public.

PR #81, opened 2022-11-23
Feat: Share Projects & Access Controls

li-xin-yi:
Add `public: Boolean` and `collaboratorIds: String[]` to the `Repo` model, enable inviting collaborators feature and public-shared repos. Probably need to run: 

```
npx prisma migrate dev --name init
```

in API container

### Updates

1. Load both your own repos and the repos where you work as a collaborator, tagged with *public* or *private*.
2. For your own private repos, show a button for sharing

image

3. Prompt a window when sharing, enter the email address, then the collaborator can get the repo in "shared repos" section.

image

4. Alert an error message when
 - The user doesn't have the privilege to share
 - No user is signed up with this email
 - The invited user is the owner or one of their collaborators invited before.
 - The repo is public

image

5. Also, an option in Right-click Menu in Repo page, no need to go back the homepage

image

### Works in the future or on the way

- Now, users can't access a private repo in which they are neither owners nor collaborators. But it won't show the error, it just keeps loading. I may fix it with explicit error message later. (ETA 1 day)
- The access feature seems not in a fine-grained. Anyone can access a public repo and edit anything in it. It's terrible. We may consider separating the read and write access individually for the public ones. Besides, even someone hasn't logged in is still supposed to read any public repos. (ETA 3 days ~ 1 week)
- For those read-only public repos, we should allow fork/clone feature from unauthenticated users. Then collaborators in a public repo can be enabled again. (ETA 1-2 weeks)
-  UI improvement, I just finished the function, but it looks ugly (It depends, please give me some ideas if you have any suggestion)

close https://github.com/codepod-io/codepod/issues/46


lihebi (approved):
Great PR, thanks Xinyi!

Issue #74, opened 2022-11-22
Pod UI polishing Nov. 22


forrestbao:
1. Move the output hide/show button to the top right (but not too right) of output box. Change icon to a simple phrase "Hide output". 
2. If output hidden, prompt the content is hidden, along with the "Show it" link. 
    ![image](https://user-images.githubusercontent.com/438579/203251370-8e3384fa-4afd-49e8-80eb-2a9926ff491c.png)
3. Drag highlight (random color shadow) is turned on when a toolbar button is clicked. Please fix this bug. 
4. Set minimal (already there) and maximal width for output box. Wrap output content if longer than maximal width. 
5. When the output box is to the right of a code box, set the minimal height of the output box to be equal to that of the code box.

Issue #77, opened 2022-11-22
Log-in page by default for users not logged in


forrestbao:
**Problem**

Currently, if a user is not logged in, s/he sees a blank dashboard. S/he has to take an extra step to click the "Login" menu to log in. 

![image](https://user-images.githubusercontent.com/438579/203264122-fe8cec9b-0f79-4b55-94c6-0bd3ba54af61.png)

**Changes to be made**

If browser has no log in info, then bring up the log-in page automatically.

li-xin-yi:
Re-assign it to me, please. I'm also fixing the unauthenticated issue mentioned in https://github.com/codepod-io/codepod/pull/81 these dayas.

PR #71, opened 2022-11-21
feat: add button to control output

nibilin33:
issue [67](https://github.com/codepod-io/codepod/issues/67)

PR #93, opened 2022-11-25
feat: refactor output box

nibilin33:
issue [74](https://github.com/codepod-io/codepod/issues/74)
1. move toggle button to outputbox
2. fix dashboard loading
3E6158D1-624A-42BC-B166-79E54B59A262

### DIFF
--- a/ui/src/pages/repos/index.tsx
+++ b/ui/src/pages/repos/index.tsx
@@ -186,10 +186,23 @@ function RepoLine({ repo, deletable, sharable }) {
   );
 }
 
-function Repos({ url = FETCH_REPOS, type = RepoTypes.repo }) {
+function Repos({
+  url = FETCH_REPOS,
+  type = RepoTypes.repo,
+  callback = (...arg) => {},
+  onLoading = (load) => {},
+}) {
   const { loading, error, data } = useQuery(url);
-  if (loading) return <p>Loading...</p>;
-  if (error) return <Alert severity="error">{error.message}</Alert>;
+  if (loading) {
+    onLoading(loading);
+    return null;
+  } else {
+    onLoading(loading);
+  }
+  if (error) {
+    callback(error.message);
+    return null;
+  }
   const repos = data[type].slice().reverse();
   return (
     <Box>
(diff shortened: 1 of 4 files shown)


ui/src/pages/repos/index.tsx
```diff
@@ -270,7 +285,27 @@ export default function Page() {
...
+      <Repos
+        callback={(arg) => {
+          if (arg) {
+            setError(arg);
+          }
+        }}
+        onLoading={(value) => {
+          setLoading(value);
+        }}
+      />
```
lihebi, line 308: What was the issue here?

lihebi (commented):
Thanks! The two things in this PR do not seem to be related. I would suggest handling only issue #71 in this PR and leaving the second "dashboard loading fix" to another PR. Also, what was the dashboard loading issue?

lihebi:
And your image link seems to be broken?

ui/src/pages/repos/index.tsx
nibilin33: there is no issue, i just find that there are two loading hint. so i fixed it in the same time

lihebi: I see, let’s remove this, because this adds extra complexity. We’ll redirect to login page anyway, so there won’t be warnings. #77

lihebi (approved):
Thanks!

ui/src/pages/repos/index.tsx
li-xin-yi: No worry. I'm taking over this part. But I want to confirm if it redirects the login page immediately or gives any hints and then jump to the login page?

nibilin33: normally, it needs show hint firstly and then redirects

ui/src/pages/repos/index.tsx
```diff
@@ -270,7 +282,20 @@ export default function Page() {
...
+            justifyContent: "center",
+          }}
+        >
+          <CircularProgress />
+        </Box>
+      )}
+      <Repos
+        onLoading={(value)=>{
+          setLoading(value)
+        }}/>
```
lihebi, line 298: FYI #106

Issue #75, opened 2022-11-22
Canvas-view polishing Nov. 22


forrestbao:
a. Turn left banner into a [modern drawer](https://mui.com/material-ui/react-drawer/) that can expand and fold in. Be sure that the drawer keeps open until the user closes. And remain closed until the user opens it. 
c. Add [breadcrumb](https://mui.com/material-ui/react-breadcrumbs/) at the top banner to display where the user is at. When in dashboard, show `Codepod/Dashboard`. When in a repo, show `Codepod/{REPO_NAME}` and be sure that `Codepod` is a hyperlink that go back to dashboard. . When in user profile page, show `Codepod/{USERNAME}'s Profile` and  be sure that `Codepod` is a hyperlink that go back to dashboard.

PR #96, opened 2022-11-26
refactor: sidebar and header

nibilin33:
issues[75](https://github.com/codepod-io/codepod/issues/75)
final effect
t-rec

### DIFF
--- a/ui/src/components/Header.tsx
+++ b/ui/src/components/Header.tsx
@@ -2,7 +2,7 @@ import { Link as ReactLink } from "react-router-dom";
 
 import { useState } from "react";
 
-import { useNavigate } from "react-router-dom";
+import { useNavigate, useLocation } from "react-router-dom";
 
 import Box from "@mui/material/Box";
 import MenuIcon from "@mui/icons-material/Menu";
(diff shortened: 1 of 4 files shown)


lihebi:
Looking great! Now the sidebar cannot be pinned. But some information are supposed to be available to the users all the time:
- the cloud saving status
- the kernel status

Screenshot 2022-11-25 at 6 46 11 PM

Let's implement the same UI for saving status as Google Docs:

Screenshot 2022-11-25 at 6 49 30 PM

And the kernel status as Colab:

Screenshot 2022-11-25 at 6 51 31 PM

@forrestbao any thoughts?

forrestbao:
let's study Deepnote's UI. It does not show whether it's saving to the cloud (because it is supposed to) nor show whether a kernel is connected on the side bar -- if it is not connected, it will show a hovered notification. I like this design actually. 

So why don't we move things that are in the side bar to the top bar? Given that the top bar is not crowded and we have only a couple of items in the side bar? 

image

lihebi:
>  nor show whether a kernel is connected on the sidebar -- if it is not connected, it will show a hovered notification.
Deepnote does always show the kernel status, both in the sticky top bar, as well as in the right bar.

I think the "saved to cloud" feedback gives users more confidence. I prefer to have it as well, just like Google Docs and Feishu Docs.

Lastly, the sidebar in this PR cannot be pinned: you cannot have it while coding. Let's change it to a pinnable sidebar instead, like [this](https://mui.com/material-ui/getting-started/templates/dashboard/).

Screenshot 2022-11-25 at 11 29 15 PM

lihebi:
To summarize, here're my requests:
- show the cloud-saving status all the time (next to the breadcrumb)
- show the kernel status all the time (on a sticky top bar?)
- make the sidebar pinnable (and make it show by default?)

nibilin33:
if these informations move to other places, it seems there is no need for sidebar?

nibilin33:
if these informations move to other places, it seems there is no need for sidebar?

lihebi:
We will have more items in the sidebar soon. So let’s keep the sidebar.

forrestbao:
So what's the status of this? I simply wanted to maximize the canvas area. It was too small on my 13-inch laptop. 

Side bar is usually foldable, holding information that is not constantly needed. 
So I suggest we move information that users need to see constantly to the top bar. 

@lihebi ?

forrestbao:
>     * make the sidebar pinnable (and make it show by default?)

@nibilin33  I talked with Hebi. He wants the sidebar to remain open until a user closes it, and I agree with it. In DeepNote, the sidebar remains there until a user closes it. 

I read the code. You used Material UI's `<Drawer  />` It actually has a persistent mode which perfectly meet Hebi's needs. 

https://mui.com/material-ui/react-drawer/#persistent-drawer

Additionally, in the demos of persistent drawer in the link above, the draw does not overlay above the "main page" nor the top bar. This is the behavior we expect.

nibilin33:
* make the sidebar pinnable (and make it show by default?)
this way needs totally rewrite,header is global share. it will take a lot time. I will just commit header breadcrams firstly

lihebi:
> make the sidebar pinnable (and make it show by default?)
> this way needs totally rewrite

The previous sidebar is pinnable and shows by default. I think what @forrestbao wants is to change the text "hide" and "show" to some icons.

Screenshot 2022-11-29 at 10 14 19 AM

lihebi:
> header is global share

But yes, I agree this needs more code and design to pass repo states to the header. Move this discussion to #105.

PR #101, opened 2022-11-28
Fix: auth error auto redirect & tree-sitter load path & toast notification

li-xin-yi:
1. When you haven't logged in, show the error message and auto redirect to login page:

image

2. When you access a repo that you are not supposed to be there (e.g., a private repo where you are neither an author or collaborator, or a wrong URL), show the error message and auto redirect to your dashboard page.

image

3. Fix the load path issue in parser, otherwise, a user can't enter a repo by URL directly. Thanks to @lihebi , but it doesn't promise the parser ready for analyzing code when you directly jump to a repo by URL. We may modify more code to support waiting for the loading in the future.

4. Fix the bug that a public repo can't be created. （But I think the UI can be improved, it looks a little misleading)

image

5. Change the notification in share project dialog into toast style, which is auto hidden is 3 seconds in the bottom-left of a page. So the success message can also be left alone for a while after auto closing the dialog.

image

image

close https://github.com/codepod-io/codepod/issues/77

### DIFF
--- a/api/src/resolver_repo.ts
+++ b/api/src/resolver_repo.ts
@@ -76,6 +76,7 @@ export async function myCollabRepos(_, __, {userId}) {
 }
 
 export async function repo(_, { id }, { userId }) {
+  // a user can only access a private repo if he is the owner or a collaborator
   const repo = await prisma.repo.findFirst({
     where: { OR: [
       { id, public: true },
--- a/api/src/resolver_repo.ts
+++ b/api/src/resolver_repo.ts
@@ -79,8 +79,8 @@ export async function repo(_, { id }, { userId }) {
   const repo = await prisma.repo.findFirst({
     where: { OR: [
       { id, public: true },
-      { id, owner: { id: userId!} },
-      { id, collaboratorIds: { has: userId!} },
+      { id, owner: { id: userId || "undefined"} },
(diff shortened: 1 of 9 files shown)


api/src/resolver_repo.ts
```diff
@@ -76,11 +76,12 @@ export async function myCollabRepos(_, __, {userId}) {
...
 
 export async function repo(_, { id }, { userId }) {
+  // a user can only access a private repo if he is the owner or a collaborator
   const repo = await prisma.repo.findFirst({
     where: { OR: [
       { id, public: true },
-      { id, owner: { id: userId!} },
-      { id, collaboratorIds: { has: userId!} },
+      { id, owner: { id: userId || "undefined"} },
+      { id, collaboratorIds: { has: userId || "undefined"} },
```
lihebi, line 84: What is the purpose of the string "undefined" here?

lihebi (approved):
Thanks!

api/src/resolver_repo.ts
li-xin-yi: When a guest visits a public repo, the `userId` could be null, but we should still return the repo. Though `OR` relationship may be short-circuited by the condition `public==true`, a null `userId` still throw a type error for the following two conditions.

Issue #91, opened 2022-11-24
Google doc style repo creation


lihebi:
Currently, one must give a name when creating a repo. This creates an extra barrier for users to get started.

On GitHub, each project has a handle, e.g. `codepod-io/codepod`. We won't use that. Instead, we use Google Docs's approach. Users can create a repo, and it will be titled "Untitled". We can have many "Untitled" repos, no need to be unique (but they will have unique hash ID). The title can be edited at any time.

For reference, this is Google Docs interface:

<img width="484" alt="Screenshot 2022-11-23 at 11 48 27 PM" src="https://user-images.githubusercontent.com/4576201/203723676-99518484-b664-4863-aa3f-332433b6b255.png">

lihebi:
@forrestbao Any input on this?

lihebi:
"If we don't have a handle like `torvalds/linux`, how should we share/discover/refer-to?" One would ask.

Easy; everyone is Youtuber here. You have a customizable homepage to showcase your work, and each of your project has a short unique URL like https://youtu.be/Npu7jkJk5nM

<img width="1176" alt="Screenshot 2022-11-23 at 11 52 51 PM" src="https://user-images.githubusercontent.com/4576201/203724791-84fd7214-7cd4-48ac-9083-79d0b9773f6d.png">

forrestbao:
I think a better analogy is Google CoLab. In Google Colab, once you create a notebook, the default name is `untilted.ipynb`. Then you can changed it as in Google Docs. 
Yeah, I like this idea. 

On a related note, we can also remove "private" switch when creating a notebook. Just let a notebook by default be private until a user chooses to share it. 

Then, on yet another related note, "share" means "public"? What about team share? We shall follow Google Docs sharing scheme

![image](https://user-images.githubusercontent.com/438579/203876269-c8f80fd7-6085-4280-9879-b5f71c8d3c6c.png)

forrestbao:
> "If we don't have a handle like `torvalds/linux`, how should we share/discover/refer-to?" One would ask.
> 
> Easy; everyone is Youtuber here. You have a customizable homepage to showcase your work, and each of your project has a short unique URL like https://youtu.be/Npu7jkJk5nM


You can take a look at how DeepNote does it. For example, 

[...]

PR #105, opened 2022-11-29
refactor:header

nibilin33:
issue [75](https://github.com/codepod-io/codepod/issues/75)
implement breadcrumb in the header to display where the user is at

### DIFF
--- a/ui/src/pages/repos/index.tsx
+++ b/ui/src/pages/repos/index.tsx
@@ -103,7 +103,7 @@ function RepoLine({ repo, deletable, sharable }) {
       sx={{ "&:last-child td, &:last-child th": { border: 0 } }}
     >
       <TableCell align="center">
-        <Link component={ReactLink} to={`/repo/${repo.id}`}>
+        <Link component={ReactLink} to={`/repo/${repo.id}?name=${repo.name}`}>
           <Box
             sx={{
               display: "flex",
--- a/ui/src/components/Header.tsx
+++ b/ui/src/components/Header.tsx
@@ -2,7 +2,7 @@ import { Link as ReactLink } from "react-router-dom";
 
 import { useState } from "react";
 
-import { useNavigate } from "react-router-dom";
+import { useNavigate, useLocation } from "react-router-dom";
 
 import Box from "@mui/material/Box";
 import MenuIcon from "@mui/icons-material/Menu";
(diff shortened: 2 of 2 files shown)


ui/src/pages/repos/index.tsx
```diff
@@ -103,7 +103,7 @@ function RepoLine({ repo, deletable, sharable }) {
       sx={{ "&:last-child td, &:last-child th": { border: 0 } }}
     >
       <TableCell align="center">
-        <Link component={ReactLink} to={`/repo/${repo.id}`}>
+        <Link component={ReactLink} to={`/repo/${repo.id}?name=${repo.name}`}>
```
lihebi, line 106: We don't want the repo name to be in the URL, it's duplicate information. Although it works for this link, a shareable link like `https://app.codepod.io/repo/repo_HwYAxxbF` won't work.  There got to be another way to pass the name to the header.

For example, we could move the `<Header>` component down into the `<Repo>` component.

nibilin33: it's normally store path breadcrums in the location
image
deepnote parse location to show the path

forrestbao: @nibilin33 
1. Please do not change the design of CodePod. If you have to, discuss with me and/or Hebi before you change. We ask you to only change the UI. If you keep doing so, we will not be able to work with you. 
2. Your change is passing an argument `name` what will be not handled by CodePod server for now. Such a change could break other parts of the system. Hence, please stop making unilateral changes. 
3. You referred to DeepNote's practice. But we have a pending discussion on who to generate URLs. Actually I pointed out that DeepNote's design is not the smartest. https://github.com/codepod-io/codepod/issues/91 

cc @lihebi

nibilin33: ?? the situation is that shareable link do not contain name info so once the user use shareable link, the header path will be not correctly but user can still enter the system.

lihebi: > shareable link do not contain name info

The link already has the repo ID; including the repo name would be redundant. Instead, the repo name should be loaded from the server and the header can then be set.

> user can still enter the system

What do you mean?

forrestbao: > > shareable link do not contain name info
> 
> The link already has the repo ID; including the repo name would be redundant. Instead, the repo name should be loaded from the server and the header can then be set.

Agree. Querying the database to get the repo name from repo id is the correct way. Repo ID is permanent and unique. Repo name can change. 

> 
> > user can still enter the system
> 
> What do you mean?

I think she means that users can still view and edit a shareable repo but the repo name will be displayed incorrectly.

nibilin33: I just explain this change didn't affect server just ui display.

nibilin33: that's it. nice talking. i'm done.

PR #115, opened 2022-12-02
Fix: read/write access control for public repos

li-xin-yi:
For a public repo, even an anonymous user can read everything in it, but only the owner and collaborators are allowed to edit. In this PR: we further distinguish the access for different roles in a repo, enforced the access rule both on the front-end and back-end side. 

- front-end: hide operation UI for guests, including auto saving button, runtime kernel in the sidebar, new code/scope menu
- back-end: prohibit mutate query when auth fails

Besides, remove the awareness info (remote cursor/selection) of guests, only show the operation update from an owner/collaborator. A guest can never edit/run code in the pod or drag a pod.

image

Now, a public repo can also be shared with collaborators.

TODO
- [ ] the resizing button and drag handles sometimes do not work properly when multiple users are collaborating on the repo.

### DIFF
--- a/api/src/resolver_repo.ts
+++ b/api/src/resolver_repo.ts
@@ -5,7 +5,14 @@ const prisma = new PrismaClient();
 
 // console.log("resolver_repo.ts", Prisma.RepoInclude);
 
-async function ensurePodAccess({ id, userId }) {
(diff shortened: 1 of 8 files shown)


lihebi (approved):
Great, thanks!

PR #117, opened 2022-12-03
fix: fix typo

pionxzh:


lihebi (approved):
Thanks! The `dev.md` is deprecated though, I need to clean up the documents a bit.

PR #120, opened 2022-12-05
Fix: Lock interactivity & sync width resize & improve render performance

li-xin-yi:
1. **Fix several node initialization issues**: specify custom drag handle, check init state after the provider synced, specify width when adding pod to the db, etc.
2. **Enforce write access control for guests**: hide the interactivity button (lock) globally, disable the resize box, and discard any node info update to the `yjs` servers from a guest. In this early stage, a user may feel a little inconvenient to select text in a pod without the button, however, they can fix the viewpoint by pressing <kbd>Control</kbd> temporarily. We (@lihebi ) will implement some other robust method to add auth layer to the yjs servers and fix the viewpoint automatically in the future if we have time.

image

3. **Sync code pod resize in real-time**: rewrite the logic of resetting the pod width, which makes all users observe the dimension update immediately when a user conduct resizing operation on it.
4. **Improve the render speed when a `codeNode` is re-rendered**: manage the `isEditorBlur` state globally instead of locally, otherwise, `MyMonaco` will be re-render over and over again any time the `codeNode` changes. If we click on the editor and then click other places, you can see it causes the re-rendering of  `MyMonaco` twice, which takes much time:

Before

But after we apply this PR,  any update propagation stops at the `MyMonaco` .

After It reduces the re-render overhead a lot, especially when you drag a codeNode.

### DIFF
--- a/ui/src/components/Canvas.tsx
+++ b/ui/src/components/Canvas.tsx
@@ -62,6 +62,7 @@ interface Props {
   data: any;
   id: string;
   isConnectable: boolean;
+
   // selected: boolean;
 }
 
@@ -77,6 +78,7 @@ const ScopeNode = memo<Props>(({ data, id, isConnectable }) => {
     translate: [0, 0],
   });
   const selected = useStore(store, (state) => state.selected);
+  const role = useStore(store, (state) => state.role);
 
   const onResize = useCallback(({ width, height, offx, offy }) => {
     const node = nodesMap.get(id);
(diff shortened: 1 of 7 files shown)


lihebi:
Added a few commits:
- There was a (seemingly unnecessary) dimension change at the very beginning of canvas page, which causes dirty status of all CodeNodes to be set. Added a check to prevent that.
- Directly set pod.dirty status without checking hash.
- Enable redux devtools for easier debugging.

lihebi (approved):
Thanks, Xinyi!

Issue #87, opened 2022-11-24
keyboard short cut to move focus/editing to a new cell


forrestbao:
Can we allow programmers to move the  focus/editing from one cell to another using keyboard shortcuts on top of mouse operations? 

In Jupyter, it was easy, just quit Edit mode into Command mode and then use up and down arrow keys to move between cells. Maybe we should simply use up, down, left, and right arrows. But we need to first  enable Edit and Command modes (#88) .

forrestbao:
@pionxzh

forrestbao:
The hopping between cells via keyboard operations has been done in https://github.com/codepod-io/codepod/pull/327 
But we still need the keyboard shortcut to switch between edit mode and command mode. https://github.com/codepod-io/codepod/issues/88

senwang86:
Please refer PR #398.

PR #129, opened 2022-12-09
feat: support naming on pod and scope

### LINKED ISSUE #97: Allowing naming pods and scopes
Add a small box at the headbar of a pod or scope where users can click and name a pod or a scope. 

This will help code management and visualization. 

What do you think? @lihebi

pionxzh:
closes #97

long text will become an ellipsis.
image

We can further improve some of the controlling behaviors around this name input field. (hint, style, arrow key navigation)

### DIFF
--- a/ui/src/components/Canvas.tsx
+++ b/ui/src/components/Canvas.tsx
@@ -24,6 +24,7 @@ import ReactFlow, {
 import "reactflow/dist/style.css";
 
 import Box from "@mui/material/Box";
+import InputBase from "@mui/material/InputBase";
 import CircularProgress from "@mui/material/CircularProgress";
 import Tooltip from "@mui/material/Tooltip";
 import IconButton from "@mui/material/IconButton";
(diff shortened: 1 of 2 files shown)


forrestbao:
@pionxzh what do you mean by hint and style? 
I love the arrow key idea.

pionxzh:
> How do we intend to manage the name state?
> 
> It seems that you only maintain the pod names during one-time visit life-cycle in this PR.
> 
> I just wonder which mode we should design
> 
> - keep the consistent pod names for all users all the time ( sync by DB and yjs server)
> - or, let each user tags the pod with their preferred name even in one same repo
> - or, reset the pod names by default after refreshing
> 
> It involves some back-end DB and yjs server works. I can work on it after we agree on either mode.

Good point. Actually I didn't check did it really get persisten to the DB or not, but it did remember my changes after reloading.

I prefer the idea of syncing pod name to DB and YJS.

lihebi (approved):
Great feature, thanks Pionxzh!

forrestbao:
@pionxzh the arrow key idea was mentioned here: https://github.com/codepod-io/codepod/issues/87 I will temporarily assign to you.

PR #132, opened 2022-12-10
Feat: Sync pod name among users in real-time

li-xin-yi:
A tiny fix on sharing the pod name data by the yjs server in real-time

1. Sync the pod name of each pod right after the name changes. (ref: https://github.com/codepod-io/codepod/pull/129#issuecomment-1344641706)
2. Prohibit a guest user from modifying pod names in public repos (as the title in gray and immutable)

image

3. Disable dragging on the pod name title area (ref: https://github.com/codepod-io/codepod/pull/129#issuecomment-1344067882)


lihebi (approved):
Awesome, thank you!

PR #134, opened 2022-12-10
don't set name on start when data.name is undefined

lihebi:
setPodName caused sync on page refresh. This PR fixes it.

### DIFF
--- a/ui/src/components/Canvas.tsx
+++ b/ui/src/components/Canvas.tsx
@@ -112,6 +112,7 @@ const ScopeNode = memo<Props>(({ data, id, isConnectable }) => {
   }, []);
 
   useEffect(() => {
+    if (!data.name) return;
     setPodName({ id, name: data.name || "" });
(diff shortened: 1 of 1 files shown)


li-xin-yi:
What if a user clear the pod name (make it empty)?

lihebi:
I think this condition only checks for undefined value but not empty string. Please see if that’s a problem, and if so let’s check data.name === pod.name as well.

li-xin-yi:
I introduced a check in previous PR (https://github.com/codepod-io/codepod/blob/main/ui/src/lib/store.tsx#L458), to mark a pod dirty, we may first check if the value is actually updated.

lihebi:
> I introduced [a check in previous PR](https://github.com/codepod-io/codepod/blob/main/ui/src/lib/store.tsx#L458), to mark a pod dirty, we may first check if the value is actually updated.

Yes, this looks good.

Issue #138, opened 2022-12-12
1. sync important fields of store.pods[id]; 2. sync between yjs and database


lihebi:
In collaborative editing, only yjs.nodesMap is synced, but the store.pods[id].xxx fields are not updated. We need to sync the store.pods, otherwise, later operations might create issues, e.g., https://github.com/codepod-io/codepod/pull/136#discussion_r1045489669.

We also need to design carefully to maintain consistency between yjs and the database.

Ref: #136

Fields:
- [ ] position (after dragging)
- [ ] parent/children (after dragging into a scope)
- [x] content

PR #140, opened 2022-12-13
Fix: sync store.pods[id].parent/position immediately on updates from anywhere

li-xin-yi:
I don't know what triggered such a horrible code reformation in `canvas`, actually this is just a very tiny modification. Eventually, I adopted another simpler optimization strategy other than [what I proposed here](https://github.com/codepod-io/codepod/pull/136#discussion_r1045877268).  Please test it carefully before merging, it works on my side, but I'm not sure if any issue exists.

A brief summary:

1. Make all tooltips not draggable.
2. Rewrite `store.setPodPosition` and `store.setPodParent`, add a parameter `dirty: bool`, set `pod.dirty = pod.dirty | dirty`, that is, only triggers remote update request when the argument `dirty=true`. Note that we use `|` (or) operation instead of `=` (assign) to avoid a dirty pod cleaned by another local `setPodPosition` or `setPodParent` with  `dirty=false` before the update is synced to the DB.
3. For `codeNode` and `scopeNode`, `useEffect` listens on the change of node position or change, call `setPodPosition` or `setPodParent` with `dirty=false` whenever the corresponding position or parent node changes. We don't need to write the change to DB for now, but we must sync `store.pods` with the nodes in canvas.
4.  `dirty=true` is only used when stopping dragging nodes (in `onNodeDragStop`, that is, we only writes the update to DB when it is caused by operations by our current user. We don't need to write update pulled from the remote yjs server to the DB.

Possible issue:

- During dragging, the position of a node changes all the time, so we keep rendering the node and calling `setPodPosition` continually, even though it doesn't change `store.pods[id].dirty` and no data is written to the DB. One solution to mitigate the update to `store`: take a look at `dragging` as well, only update `store.pods[id]` after the dragging stops, like:

```ts
function codeNode({..., xPos, yPos, dragging}) {
// ....
useEffect( () => {
 if(!dragging) {
 let [x, y] = relativePostion(xPos, yPos); // transform XY coordinate
 setPodPosition({id, x, y, dirty: false})
}
}, [...,xPos, yPos, dragging])
// ....
}
```

But I think the performance of dragging is acceptable now, if you feel uncomfortable with the performance in the future, the solution is just for reference.

close https://github.com/codepod-io/codepod/issues/138

### DIFF
--- a/ui/src/components/Canvas.tsx
+++ b/ui/src/components/Canvas.tsx
@@ -64,211 +64,220 @@ interface Props {
   id: string;
   isConnectable: boolean;
   selected: boolean;
+  xPos: number;
+  yPos: number;
 }
 
-const ScopeNode = memo<Props>(({ data, id, isConnectable, selected }) => {
-  // add resize to the node
-  const ref = useRef(null);
-  const store = useContext(RepoContext);
-  if (!store) throw new Error("Missing BearContext.Provider in the tree");
-  const flow = useReactFlow();
-  const setPodName = useStore(store, (state) => state.setPodName);
-  const updatePod = useStore(store, (state) => state.updatePod);
-  const [target, setTarget] = React.useState<any>();
-  const nodesMap = useStore(store, (state) => state.ydoc.getMap<Node>("pods"));
-  const [frame] = React.useState({
-    translate: [0, 0],
-  });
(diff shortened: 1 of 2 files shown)


lihebi (approved):
Thanks!

Issue #99, opened 2022-11-26
Cannot move pod out of a scope


forrestbao:
Currently we cannot move a pod/scope out of a scope to dissociate a child pod/scope from a parent scope. This is very inconvenient and it looks stupid when we demo to investors.

lihebi:
First, the current pod-cannot-be-moved-out behavior is useful (i.e., the [`extent: 'parent'`](https://reactflow.dev/docs/api/nodes/node-options/#extent), e.g., the [node B.1 in this example](https://reactflow.dev/docs/examples/layout/sub-flows/)).

We do want to move a pod out. We could probably add a "release button", which, once clicked, temporarily removes the `extent: 'parent'` field to allow it to be moved out.

sudongyuer:
if we just removes the `extent: 'parent' ` , the node's parent still exist , shouldn't we consider removing its parent node?

sudongyuer:
I have already change the pr ( use a button to control the extent option )  #143

PR #143, opened 2022-12-13
feat: support remove scope

sudongyuer:
#99


ui/src/components/Canvas.tsx
```diff
@@ -520,6 +523,19 @@ const CodeNode = memo<Props>(function ({
...
+  useEffect(() => {
+    // get relative position
+    const node = nodesMap.get(id);
+    const parentNode = nodesMap.get(node?.parentNode??'');
+    // console.log(node, parentNode)
+    if(node?.parentNode && !node.position.x && dragging){
+      node.parentNode = undefined
+      node.data!.parent = 'ROOT'
+      setPodParent({ id, parent: 'ROOT', dirty: false });
+    }
```
lihebi, line 535: Two issues:
1. this allows only dragging a pod out of a scope from **left** side, but not other 3 sides. This behavior sounds inconsistent.
2. we need to update the position (which is relative to the parent scope), otherwise, the pod will "jump" upon moving out.

More broadly speaking, we probably want another way to dissociate the pod from the scope, other than simply dragging, because the current pod-cannot-be-moved-out behavior is useful. We could probably add a "release button" which, once clicked, temporarily remove `extent: 'parent'` field to allow it to be moved out.

ui/src/components/Canvas.tsx
```diff
@@ -1118,24 +1134,24 @@ export function Canvas() {
...
+
+            {/*    return "#d6dee6";*/}
+            {/*  }}*/}
+            {/*  nodeColor={(n) => {*/}
+            {/*    if (n.style?.backgroundColor) return n.style.backgroundColor;*/}
+
+            {/*    return "#f4f6f8";*/}
+            {/*  }}*/}
+            {/*  nodeBorderRadius={2}*/}
+            {/*/>*/}
```
lihebi, line 1151: Why do you remove the minimap?

ui/src/components/Canvas.tsx
```diff
@@ -1118,24 +1134,24 @@ export function Canvas() {
...
+            {/*<Controls showInteractive={role !== RoleType.GUEST} />*/}
+
+            {/*<Background />*/}
```
lihebi, line 1154: and the controls and the canvas backgrounds?

lihebi (commented):
Thanks, sudongyuer! There appear to be a few coding and design issues.

ui/src/components/Canvas.tsx
sudongyuer: > Thanks, sudongyuer! There appear to be a few coding and design issues.

Yes , so i make this pr to Draft

ui/src/components/Canvas.tsx
sudongyuer: Because i don't know many context about the source code , i just remove them for debug , when i make this pr ready for review , i will reset them

lihebi: Ah, I see, sounds good!

ui/src/components/Canvas.tsx
lihebi: ditto

sudongyuer:
This pr still have a issue that i don't know how to fix :
- When toggle the extent button , it can't immediate render

sudongyuer:
This is this PR's behavior
https://user-images.githubusercontent.com/76603360/207765493-a8b937d3-4067-46f2-b007-1092d7ed69b8.mov

forrestbao:
@sudongyuer What will happen if you move the scope after the code cell has been moved out of it? Is the parent of the cell updated to the scope at the outer layer?

sudongyuer:
> @sudongyuer What will happen if you move the scope after the code cell has been moved out of it? Is the parent of the cell updated to the scope at the outer layer?


https://user-images.githubusercontent.com/76603360/207772790-c79dd567-371c-485f-9e23-4770c4284563.mov

sudongyuer:
You can see more contexts in #99 
image

sudongyuer:
> 

The node's parent will not change , it will remain the relationship before

forrestbao:
@sudongyuer yes, i saw the discussion in #99. Should we update the parent then? Otherwise, the code scoping is not correct.

sudongyuer:
> 

 

> @sudongyuer yes, i saw the discussion in #99. Should we update the parent then? Otherwise, the code scoping is not correct.

I just think this pr is to control the extent option in node , if you want to remove node's parent relationship , should be better add another feature  to support that . Also need some suggestions from @lihebi

lihebi:
Controlling only the "extent" option itself doesn't make a useful feature. We should implement the logic for updating node parent relations in this PR.

sudongyuer:
https://user-images.githubusercontent.com/76603360/208028198-6ccbb229-1f9d-4ac0-8cfd-7ad0934b2f21.mov

li-xin-yi:
Thanks for the great PR!

Please take care when updating the parent, you should update nodesMap[id].level recursively as well. My previous PR for reference: https://github.com/codepod-io/codepod/pull/145

forrestbao:
I think there can be an easy solution. After a user clicks the "remove extent" button (the name needs to be changed later into something like "change scope") of a cell/scope, the cell/scope will move with the mouse. The user can move the mouse to move the cell/scope to where s/he sees fit and drop it there by clicking the mouse again. Just detect the node on which the mouse click happens and let it be the parent of the cell/scope just moved.

sudongyuer:
> Is recursive updating necessary? After a user clicks the "remove extent" button (the name needs to be changed later into something like "change scope") of a cell/scope, the cell/scope will move with the mouse. The user can move the mouse to move the cell/scope to where s/he sees fit and drop it there by clicking the mouse again. Just detect the node on which the mouse click happens and let it be the parent of the cell/scope just moved.

Now when click the `remove from scope ` button , it will just remove the parent relationship , it's not save the relationship before , so when you click again , it will not restore the previous parent-son state , so recursive update is necessary now  .

li-xin-yi:
> Is recursive updating necessary? After a user clicks the "remove extent" button (the name needs to be changed later into something like "change scope") of a cell/scope, the cell/scope will move with the mouse. The user can move the mouse to move the cell/scope to where s/he sees fit and drop it there by clicking the mouse again. Just detect the node on which the mouse click happens and let it be the parent of the cell/scope just moved. 

It's always necessary to update the levels of all involved nodes when any scope relationship changes. The level determines the pod color in the mini-map, we need to show everything in the mini-map correctly in time.

The idea to drag and release sounds cool. But I want to make sure about something related to my implementation: what will a collaborator see during the process of changing scope? the complete movement track or nothing until it finishes?

lihebi:
> Screen.Recording.2022-12-16.at.13.20.56.mov

Thanks!

When you click the release button, the pod "jumps" out. It is because you updated its parent but didn't update its position (which is relative to the parent). We don't want that jump.

lihebi:
> [xinyi] you should update nodesMap[id].level recursively as well
> [forrest] Is recursive updating necessary?
> [xinyi] It's always necessary to update the levels

Xinyi had explained it. Just an echo: Xinyi was talking about recursively updating the **levels** property of descendant nodes, which is missing in this PR.

> [forrest] After a user clicks the "remove extent" button of a cell/scope, the cell/scope will move with the mouse.

To make things simple, we could just drop the node to ROOT scope when the release button is clicked. We probably want to show some indicators (e.g., a highlighted shadow) to mark it global. Then we can drag-n-drop to the location.

> [xinyi] what will a collaborator see during the process of changing scope?

This will no longer be a problem if we adopt the "drop-to-ROOT-scope" approach above.

sudongyuer:
> > [xinyi] you should update nodesMap[id].level recursively as well
> > [forrest] Is recursive updating necessary?
> > [xinyi] It's always necessary to update the levels
> 
> Xinyi had explained it. Just an echo: Xinyi was talking about recursively updating the "**levels**" property of descendant nodes, which is missing in this PR.
> 
> > [forrest] After a user clicks the "remove extent" button of a cell/scope, the cell/scope will move with the mouse.
> 
> To make things simple, we could just drop the node to ROOT scope when the release button is clicked. We probably want to show some indicators (e.g., a highlighted shadow) to mark it global. Then we can drag-n-drop to the location.
> 
> > [xinyi] what will a collaborator see during the process of changing scope?
> 
> This will no longer be a problem if we adopt the "drop-to-ROOT-scope" approach above.

ui/src/components/Canvas.tsx
```diff
@@ -494,6 +491,27 @@ const CodeNode = memo<Props>(function ({
...
+    }
+  }
+
+  const removeParentByNodeById = (id) => {
+    const node = nodesMap.get(id)
+    if(node?.parentNode){
+      node.parentNode = undefined
+      node.data!.parent = 'ROOT'
+      nodesMap.set(id,node)
+      updateLevel(node.id, (node as any).level - 1);
```
sudongyuer, line 511: This line update the level

lihebi: Yeah, thanks! I posted that comment to just re-answer Dr. Bao’s original question.

sudongyuer:
Thx xinyi ~

Issue #144, opened 2022-12-13
The levels of nodes inside a scope is not updated when dragging the scope into a larger scope


li-xin-yi:
### Bug reproduce:

![image](https://user-images.githubusercontent.com/10118462/207449806-fdbaded5-dca4-4d7e-b587-2f11d5aa90a6.png)

1. create a code 1 and scope 1 individually on the canvas.
2. drag code 1 into the scope 1. By now, everything works well.
3. create a larger scope 2, drag scope 1 into scope 2

Now, the level of code 1 and scope 1 are the same. To compare that clearly, create a code 2 and drag it into scope 1, see their background colors in the minimap, which are supposed to be the same as their parents are scope 1.

### Solution

Therefore, I propose to update `node.level` in a pre-order DFS way to ensure all involved descendant nodes are in correct levels, or the rendering order and background colors are incorrect.

PR #145, opened 2022-12-13
Fix: update all children node levels when dragging into a scope

li-xin-yi:
close https://github.com/codepod-io/codepod/issues/144


lihebi (approved):
Thanks!

PR #130, opened 2022-12-09
fix: #98

fundon:
Color taken from from https://mui.com/material-ui/customization/color/#main-content.

Screenshot 2022-12-09 at 10 04 40 PM

Screenshot 2022-12-09 at 10 02 21 PM

PR #153, opened 2022-12-16
fix(ui): remove bg and circle button on scope's bar

fundon:
#130 

I think the whole canvas is too complicated to modify UI style and handle business logic at the same time.
It would be better to provide a right sidebar like the design tool, so that when you select scope or node, you can modify their styles or attributes.

It may also need a designer. :)

### DIFF
--- a/ui/src/components/Canvas.tsx
+++ b/ui/src/components/Canvas.tsx
@@ -556,10 +556,10 @@ const CodeNode = memo<Props>(function ({
         borderColor: pod.ispublic
           ? "green"
           : selected
-            ? "#003c8f"
-            : !isPodFocused
-              ? "#d6dee6"
-              : "#5e92f3",
+          ? "#003c8f"
+          : !isPodFocused
+          ? "#d6dee6"
+          : "#5e92f3",
       }}
     >
       <Handle
@@ -944,7 +944,7 @@ export function Canvas() {
    */
 
   // FIXME: add awareness info when dragging
-  const onNodeDragStart = () => { };
+  const onNodeDragStart = () => {};
 
   const onNodeDragStop = useCallback(
     // handle nodes list as multiple nodes can be dragged together at once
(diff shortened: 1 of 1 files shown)


ui/src/components/Canvas.tsx
```diff
@@ -970,10 +970,10 @@ export function Canvas() {
           if (
             mousePos.x < currentParent.positionAbsolute!.x ||
             mousePos.x >
-              currentParent.positionAbsolute!.x + currentParent.width! ||
+            currentParent.positionAbsolute!.x + currentParent.width! ||
             mousePos.y < currentParent.positionAbsolute!.y ||
             mousePos.y >
-              currentParent.positionAbsolute!.y + currentParent.height!
+            currentParent.positionAbsolute!.y + currentParent.height!
```
lihebi, line 976: We seem to be still on a different formatter (we used [prettier](https://marketplace.visualstudio.com/items?itemName=esbenp.prettier-vscode)). Can you double check?

forrestbao:
> It would be better to provide a right sidebar like the design tool, so that when you select scope or node, you can modify their styles or attributes.

This means changing the habit of users by quite a lot. Currently Google Colab, DeepNote, Noteable, and Hex all use the hover tool bar/strip design for each cell.

ui/src/components/Canvas.tsx
fundon: A configuration file may be needed in ui's fold.

fundon:
My thoughts and suggestions are mentioned above.

A right sidebar or a hover tool bar/strip would improve the experience.

lihebi (approved):
Thanks!

ui/src/components/Canvas.tsx
lihebi: Thanks for the suggestions! Since we have nothing to customize prettier, we just left it empty (which should apply the default options). Do you think this would work or not? Or did you use prettier but got this different formatting?

PR #154, opened 2022-12-16
support scoped variable via proper compile pass

lihebi:
Scoped-variable is now implemented via a proper compile pass. It is enabled by default. But it is not stable enough, you can toggle it off in the sidebar.

The following visual appearance is added to the editor:
- function definition
- (global) variable definition
- function call site
- variable use
- unbound variables

Screenshot:

Screenshot 2022-12-16 at 3 38 33 PM


lihebi:
UPDATE: hide annotations by default, and add a "show annotations" setting to toggle it.
Screenshot 2022-12-19 at 10 56 10 AM

Issue #155, opened 2022-12-18
Position to paste cells in `ctrl`-`v` way


li-xin-yi:
It's hard to implement a paste-button in the right-click menu because modern browsers restrict the read-access to system clipboard unless user explicitly does that. I tried Jupyterlab/Jupyter notebook/Colab/Deepnote, none of them implements such a paste button across pages.

![image](https://user-images.githubusercontent.com/10118462/208291180-a415ff39-383c-4bd1-912b-ae08a8bf3dd5.png)

( A hack way (custom web MIME to serialize pod data) can be used here if insisted)

`ctrl-v` way is much easier for me to implement, and faster for users to operate.  But different from those products above, which can simply appended the pasted cell to the end of the doc or cursor in 1-D space, we need a much cleared 2-D position to put the pasted cell by default, which one should we apply?

- [ ] the current center of the canvas
- [ ] last clicked position on the pane
- [ ] below/beside the copied position (not applicable for pods copied from other pages)
- [ ] others

li-xin-yi:
@lihebi @forrestbao 

I implemented the single code node copy-paste across repos feature in this experimental branch: https://github.com/li-xin-yi/codepod-1/tree/experimental/clipboard, tests and suggestions are welcome. This is just a simple demo to show my idea, even many `console.log` calls are remained to show what in your clipboard in each stage.

How to use it:

1. Click on the copy button on the code node you want to copy.
2. Now, if you choose to paste it on a text editor (e.g. VS code, chat input box in telegram, or any text input field), only the code itself will be pasted.
3. Now, if you choose to paste it on a Codepod canvas (no matter the current page or any other repo page, you can even use different browsers with different accounts), you can:
  - open the right-click menu, click on the paste button, the identical pod shows up in place.
  - or, make sure the pane is focused, then press `Ctrl` + `V`,   the identical pod appears in the center of current canvas view.

What is copied into the canvas:

![image](https://user-images.githubusercontent.com/10118462/208371799-64db857c-55dc-4044-8e63-65708a8c0daf.png)

1. Code content
2. Pod width
3. Pod name
4. Result/Error message/stdout (if the pod is executed), but the collaborator online can't see the output, because we keep runtime info independently.

How it works:

[...]

PR #158, opened 2022-12-20
Feat: cross-page single-pod copy-paste in mouse-bound `ctrl`-`v` way

li-xin-yi:
[Done] This PR is finished, ready to review now.

(Only support one single code pod by now, `Ctrl-C` is not supported yet)

Related to: https://github.com/codepod-io/codepod/issues/155

I gave up `clipboard.read/write`, removed the paste option in right-click menu, tested on Chrome, Edges, and Firefox, all works well.

How to use:
1. Copy a pod (either in this page or other pages) via the copy button
2. Click (or right-click) on the canvas (pane) once to make sure the pane is focused
3. Press `Ctrl` -`V`, a duplicated pod shows up in the center of the canvas (FIXME: because the cursor may not be in the canvas when pasting, so I just choose the center as the initial position, we can discuss the detailed rule for the initial position later)
4. Move your mouse inside the canvas, the half-transparent pasted pod will move together, but the collaborator will never see this half-transparent pod. (To be exactly, the top-left corner position of this pod comes following the mouse, we may discuss the anchor point later, which point do you prefer? The center of the pod? Or the top-left?)
5. Click on the destination position, then the pasted pod completely shows up, others can also see it on their pages.

~~Don't merge, just a demo to convey my idea and for test usages from others. The code is too dirty now, if you agree on this idea, I can continue to improve the implementation.~~

### DIFF
--- a/ui/src/components/Canvas.tsx
+++ b/ui/src/components/Canvas.tsx
@@ -645,6 +645,15 @@ const CodeNode = memo<Props>(function ({
             justifyContent: "center",
           }}
           className="nodrag"
+          onClick={(e) => {
+            const pane = document.getElementsByClassName(
+              "react-flow__pane"
+            )[0] as HTMLElement;
+            if (pane) {
+              pane.tabIndex = 0;
+              pane.focus();
+            }
+          }}
         >
           {role !== RoleType.GUEST && (
             <Tooltip title="Run (shift-enter)">
@@ -1181,7 +1190,7 @@ export function Canvas() {
         stdout: pod.stdout,
         result: pod.result,
         name: pod.name,
-        dirty: true,
+        dirty: false,
       });
 
       nodesMap.set(id, newNode as any);
@@ -1196,7 +1205,10 @@ export function Canvas() {
     };
     const handlePaste = (event) => {
       console.log("paste", event);
-      if (document.activeElement?.className !== "react-flow__pane") {
+      if (
+        event.target?.className !== "react-flow__pane" &&
+        document.activeElement?.className !== "react-flow__pane"
+      ) {
         return;
       }
       const playload = event.clipboardData.getData("application/json");
@@ -1250,21 +1262,49 @@ export function Canvas() {
           parent: node.data?.parent,
         },
       };
+      const pod = getPod(pasting);
       nodesMap.delete(pasting);
+      addPod(apolloClient, {
+        ...pod,
+        dirty: true,
+      } as any);
       nodesMap.set(pasting, newNode);
+      const currentNode = reactFlowInstance.getNode(pasting);
+      onNodeDragStop(event, currentNode, [currentNode]);
+      setPasting(null);
+    };
+    const keyDown = (event) => {
+      console.log("escapeDown", event);
+      if (event.key !== "Escape") return;
+      event.preventDefault();
+      // delete the temporary node
+      nodesMap.delete(pasting);
+      // delete the temporary pod in local store
+      deletePod(null, { id: pasting, toDelete: [] });
       setPasting(null);
     };
     reactFlowWrapper.current.addEventListener("mousemove", mouseMove);
     reactFlowWrapper.current.addEventListener("click", mouseClick);
+    document.addEventListener("keydown", keyDown);
     return () => {
       reactFlowWrapper.current.removeEventListener("mousemove", mouseMove);
       reactFlowWrapper.current.removeEventListener("click", mouseClick);
+      document.removeEventListener("keydown", keyDown);
     };
-  }, [pasting, reactFlowWrapper, setPasting]);
+  }, [
+    pasting,
+    reactFlowWrapper,
+    setPasting,
+    getPod,
+    deletePod,
+    addPod,
+    apolloClient,
+    reactFlowInstance,
+    nodesMap,
+  ]);
 
   const onPaneClick = (event) => {
     console.log("onPaneClick", event);
-    event.preventDefault();
     event.target.tabIndex = 0;
   };
 
--- a/ui/src/components/Canvas.tsx
+++ b/ui/src/components/Canvas.tsx
@@ -33,11 +33,12 @@ import Stack from "@mui/material/Stack";
 import Button from "@mui/material/Button";
 import CircleIcon from "@mui/icons-material/Circle";
 import CheckCircleIcon from "@mui/icons-material/CheckCircle";
+import ContentCopyIcon from "@mui/icons-material/ContentCopy";
 import Grid from "@mui/material/Grid";
 import PlayCircleOutlineIcon from "@mui/icons-material/PlayCircleOutline";
 import DeleteIcon from "@mui/icons-material/Delete";
 import ViewComfyIcon from "@mui/icons-material/ViewComfy";
-
+import { CopyToClipboard } from "react-copy-to-clipboard";
 import Moveable from "react-moveable";
 import { ResizableBox } from "react-resizable";
 import Ansi from "ansi-to-react";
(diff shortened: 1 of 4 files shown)


ui/src/components/Canvas.tsx
```diff
@@ -1090,17 +1123,150 @@ export function Canvas() {
...
+          clientId,
+        },
+        level: 0,
+        extent: "parent",
+        parentNode: undefined,
+        dragHandle: ".custom-drag-handle",
+        style,
+      };
+
+      addPod(apolloClient, {
```
lihebi, line 1170: In the future, we can defer the write to DB until the final drop of the node (when we allow a user to press Esc to cancel the paste). No need to implement that because that seems to require non-trivial logic. Add a TODO comment is good enough.

ui/src/components/Canvas.tsx
```diff
@@ -1090,17 +1123,150 @@ export function Canvas() {
...
+        content: pod.content,
+        error: pod.error,
+        stdout: pod.stdout,
+        result: pod.result,
+        name: pod.name,
+        dirty: true,
+      });
+
+      nodesMap.set(id, newNode as any);
+      setPasting(id);
```
lihebi, line 1188: If we set nodesMap, it would automatically sync with collaborators, right? UPDATE resolved; I saw that you checked `clientID` in various places in `lib/nodes.tsx`. That's a clever trick, looks good to me!

ui/src/components/Canvas.tsx
lihebi: Also, since we called addPod in `observer.YMapEvent.changes`:

https://github.com/codepod-io/codepod/blob/a67b5c8d0703ca6ae1d361c5ca19a7d4f0d10bd6/ui/src/lib/nodes.tsx#L106-L111

I think all `addPod` calls in `Canvas.tsx` are duplicate, right? Should we remove them? I.e. this line (Line 1170 in this PR) and the following line (Line 901 in main):

https://github.com/codepod-io/codepod/blob/a67b5c8d0703ca6ae1d361c5ca19a7d4f0d10bd6/ui/src/components/Canvas.tsx#L901

lihebi:
> 3. because the cursor may not be in the canvas when pasting, so I just choose the center as the initial position

LGTM

> 4. the top-left corner position of this pod comes following the mouse

LGTM

lihebi:
> I gave up clipboard.read/write, removed the paste option in right-click menu, tested on Chrome, Edges, and Firefox, all works well.

LGTM. Let's forget about the right-click context menu for now.

ui/src/components/Canvas.tsx
lihebi: > addPod calls in Canvas.tsx are duplicate, right?

That depends on the behavior of [`ymap.observe`](https://docs.yjs.dev/api/shared-types/y.map#observing-changes-y.mapevent). It is triggered if "other users" make the change. Would it get triggered as well if the "local user" makes the change?

li-xin-yi: > In the future, we can defer the write to DB until the final drop of the node 

In fact, when a user press the shortcut, the half-transparent node is immediately added to the pods array in local store, and nodesMap, but have not synced to the DB (see the `addpod` arguments, I set the `dirty` false, `doRemoteAdd` will not be executed.

When the user determines the destination by clicking, I delete the temporary transparent node in `nodesMap`, and add a node with everything identical but `opacity` and `clientId` removed, then write it formally to the DB.

`[ymap.observe](https://docs.yjs.dev/api/shared-types/y.map#observing-changes-y.mapevent).` on adding events here serves like the way in which we use `useEffect` to sync `pod.name` and `pod.position` before, only sync the data in `nodesMap` with `store.pods` on each user's own side.

To avoid the duplicated pod adding in local store, I check if the pod is already in `store.pods` here: https://github.com/li-xin-yi/codepod-1/blob/experimental%2Fpaste-on-canvas/ui/src/lib/nodes.tsx#L114

ui/src/components/Canvas.tsx
li-xin-yi: Yes, it is an adding event in `nodesMap`, [`ymap.observe`](https://docs.yjs.dev/api/shared-types/y.map#observing-changes-y.mapevent) triggers a `addPod` calls without remote DB writing, just adds it to the `store.pods` in each user's side. In this case, since I add the checking statement for `clientId`, collaborators will ignore it and won't call `addPod` until we delete the temporary moving node and re-add the formal node.

ui/src/components/Canvas.tsx
```diff
@@ -1090,17 +1123,150 @@ export function Canvas() {
...
+      const position = reactFlowInstance.project({
+        x: event.clientX - reactFlowBounds.left,
+        y: event.clientY - reactFlowBounds.top,
+      });
+      const node = nodesMap.get(pasting);
+      if (!node) return;
+      node.position = position;
+      nodesMap.set(pasting, node);
+    };
+    const mouseClick = (event) => {
```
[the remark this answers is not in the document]
li-xin-yi: I agree, sounds great, I will work on this feature later.

ui/src/components/Canvas.tsx
li-xin-yi: Sorry, my bad, I noticed that I set the `dirty=true` for the temporary pod adding just now. I will fix it.

li-xin-yi:
Update:

1. Fix the immediate adding to DB issue (see [this comment](https://github.com/codepod-io/codepod/pull/158#discussion_r1053832962))
2. Add `onClick` listener to tooltip of codeNode to set the pane focus, then you can paste right after clicking on the copy button without additional clicking on the pane.
3. Press `Esc` to cancel the in-progress pasting.
4. If the destination is located inside a scope, drop this codeNode into the scope like what we did in `onNodeDragStop`.
5. If there is already data on the yjs server for this code editor, reversely initialize `pod.content` to current code on when mounting the Monaco editor. Because collaborators pulled new node info from `nodesMap`, in which nodes don't have the information of code content, the `store.pods[id].content` will be initialized as null and kept until the code changes. This is a  horrible data inconsistent issue between front-end code display <-> `store.pods[id].content` <-> `pods[id].content` in DB, for example, if you copy and paste a pod in canvas, everything looks normal, but actually the new pasted pod has an empty pod.content field, if you copy this new pod and then paste, you can see nothing on the new pasted pod.

Code still needs cleaning up and performance improvement.

li-xin-yi:
Update:

- The reverse initialization of temporary pod make the pod dirty, but it won't have a truth source in DB, so the cloud uploading must fail with error. Fix it by initialize the content without setting the pod dirty.

To-do:
1. clean up the code
2. improve the performance
3. after pressing `esc`, a black inner block shows up in the pane, indicating the whole pane is selected. The event is triggered by `esc` by default (see [this source code](https://github.com/wbkd/react-flow/blob/769261fead0dbfd11f2c327787b18ffb925fc71f/packages/core/src/utils/index.ts#L66), the enter and space have the same effect on pane). I have to figure out a solution to prevent or un-select it.

lihebi:
Thanks, all look and work great!

li-xin-yi:
Updata:

Clean up the code

1. Disable duplicated pasting when a pasting is not finished.
2. Fix the tab selection black border issue [there](https://github.com/codepod-io/codepod/pull/158#issuecomment-1360811799)
3. Unselect all pods when a pasting starts, make sure the processing pods always in the front-most, never shadowed by others.

lihebi (approved):
Awesome, thanks!

ui/src/components/Canvas.tsx
```diff
@@ -907,6 +957,7 @@ export function Canvas() {
         y: position.y,
         width: style.width,
         height: style.height,
+        dirty: true,
```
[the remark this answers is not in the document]
li-xin-yi: The `dirty` parameter is removed in [`store.addPod`](https://github.com/li-xin-yi/codepod-1/blob/b8e4f75d14329b4356f7977eb1187f8682266093/ui/src/lib/store.tsx#L306) but somehow remains here. It's a bug but shouldn't have any effect on `addPod`. Can you show me how to reproduce the error?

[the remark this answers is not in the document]
lihebi: It does not reproduce on localhost. If you go to app.codepod.io, creating pods would sometime fail to sync.

But never mind, I've fixed it.

PR #160, opened 2022-12-21
feat: add rich-text node

lihebi:
This PR adds a rich-text editor using [Remirror](https://github.com/remirror/remirror), a descendant of [prosemirror](https://prosemirror.net/). You can create it via right-click menu.

TODOs:
- [x] save to DB and sync with yjs

Screenshot:

Screenshot 2022-12-21 at 9 58 38 AM


lihebi:
UPDATE: now the content is saved to DB. This rich-text editor is not synced with collaborators. The yjs integration will be handled in another PR.

lihebi:
Know issue: the app will throw the following error during development. It seems to be caused by Remirror's use of `:first-child`, ref: https://github.com/remirror/remirror/issues/1785

Screenshot 2022-12-21 at 3 12 26 PM

PR #165, opened 2022-12-23
Feat: single-pod cut-n-paste support

li-xin-yi:
~~Don't merge now, this is just a demo to align the idea of cut-n-paste functionality and tests. The code is too dirty now, clean-up is required after we agree on the idea.~~

Update: clean-up & ready for review.

How to use:

1. Click on the cut button of a pod, then the pod become half-transparent immediately.
2. This half-transparent pod will follow your mouse like pasting behavior.
3. Choose a destination location, click on it, the new pasted pod is located here and become solid.
4. If you don't want to proceed anymore, press <kbd>Esc</kbd>, the half-transparent pod disappears and the origin pod shows up again.

Note:

1. Like copying, the data will be written in your clipboard, you can still paste it into any repos even if your cutting procedure ends, until you copy new contents.
2. During the whole cut-n-paste procedure, that is, before canceled by <kbd>Esc</kbd> or confirmed by clicking, collaborators   won't observe any clues. They can't see the half-transparent pods, the origin pod remains on their side.

Any test and suggestion are welcome.

### DIFF
--- a/ui/src/components/Canvas.tsx
+++ b/ui/src/components/Canvas.tsx
@@ -910,7 +910,9 @@ export function Canvas() {
     } else {
       provider.once("synced", init);
     }
-    // check if the nodesMap on the websocket has already been initialized with node info
+
+    // cancel in-progress pasting when exiting the canvas
+    return cancelPaste;
 
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [provider]);
@@ -1284,6 +1286,7 @@ export function Canvas() {
         reactFlowBounds.width / 2,
         reactFlowBounds.height / 2,
       ];
+
       const position = reactFlowInstance.project({ x: posX, y: posY });
       position.x = (position.x - pod.width! / 2) as number;
       position.y = (position.y - (pod.height ?? 0) / 2) as number;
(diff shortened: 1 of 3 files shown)


lihebi:
Thanks, Xinyi! Yes, this is what we want.

@forrestbao FYI, this would complete #99 and deprecate #143

Later, we'll have another PR to support copy/cut for a scope (recursively for all descendants).

li-xin-yi:
Update:

Check if there is an in-progress pasting before exiting. If there is, do what as <kbd> Esc </kbd> works.

lihebi (approved):
Thank you!

ui/src/components/Canvas.tsx
```diff
@@ -1344,12 +1394,16 @@ export function Canvas() {
       checkNodesEndLocation(event, [currentNode], "ROOT");
       //clear the pasting state
       setPasting(null);
+      // delete the original (hidden) node
+      if (cutting) {
+        reactFlowInstance.deleteElements({ nodes: [{ id: cutting }] });
```
lihebi, line 1399: When cutting, once the user left-click to drop the node, it will trigger:
1. add the new node (the `addPod(apolloClient, ...)` above)
2. delete the original hidden node.

I'm having some trouble with 2. This line seems to trigger `deletePod(null, ...)` in `nodes.tsx`, so there won't be a remote `deletePod` graphql query. But I do see a `deletePod` query. Can you point out where does that `deletePod` query come from?

lihebi: I see. It should come from `onNodesDelete`.

li-xin-yi: Yes, it triggers onNodeChanges's `delete` type.

I generalize the logic in https://github.com/codepod-io/codepod/pull/184. This PR is almost done except I need a better implementation to create a scope as well as its children in DB at once. You can try that PR to see if it is the expected behavior of scope cut-copy-paste.

lihebi: Ah, I see, thanks. Yeah, that PR is a great reference on this. I'll take a close look.

PR #170, opened 2022-12-24
[WIP] generate a sharable image/png when copying

li-xin-yi:
I use this API of [carbon](https://carbon.now.sh/): https://github.com/petersolopov/carbonara

Notice that `localhost` doesn't allow [CROS](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS), we can't send a request of `application/json` payload to `localhost:3099/api/cook`, the actual request body is in `text/plain` format. So you may need to skip the content type checking by deleting this line https://github.com/petersolopov/carbonara/blob/master/src/utils.js#L118, and then run the deployment command of this api by:

```
git clone https://github.com/petersolopov/carbonara
cd carbonara
docker build -t local/carbonara .
docker run -p 3099:3000 local/carbonara -it
```

To try if the api running normally, run a `cURL` command:

```sh
curl -L http://127.0.0.1:3099/api/cook \
-X POST \
-H 'Content-Type: application/json' \
-d '{
        "code": "sum(range(114))"
    }' \
> test.png
```

image

Now, the fetch request can be viewed with correct preview in devtools, I just wonder how to take it out.

image

### DIFF
--- a/ui/src/components/Canvas.tsx
+++ b/ui/src/components/Canvas.tsx
@@ -531,18 +531,35 @@ const CodeNode = memo<Props>(function ({
     }
   }, [data.parent, setPodParent, id]);
 
+  const getPicture = async (pod) => {
+    const requesOptions = {
+      method: "POST",
+      headers: { "Content-Type": "application/json" },
+      body: JSON.stringify({ code: pod.content }),
+      mode: "no-cors" as RequestMode,
+      accept: "image/png",
+    };
+    const res = await fetch("http://localhost:3099/api/cook", requesOptions);
+    const pic = await res.blob();
+    // navigator.clipboard.write([new ClipboardItem({ "image/png": pic })]);
+  };
+
   const onCopy = useCallback(
-    (clipboardData: any) => {
+    async (clipboardData: any) => {
       const pod = getPod(id);
       if (!pod) return;
-      clipboardData.setData("text/plain", pod.content);
-      clipboardData.setData(
-        "application/json",
-        JSON.stringify({
-          type: "pod",
-          data: pod,
-        })
-      );
+
+      clipboardData.setData("text/plain", "");
+      // clipboardData.clearData();
+      // clipboardData.setData(
+      //   "application/json",
+      //   JSON.stringify({
+      //     type: "pod",
+      //     data: pod,
+      //   })
+      // );
+      getPicture(pod);
+      console.log("copied", clipboardData);
     },
     [getPod, id]
   );


li-xin-yi:
The API even works well with Python requests:

image

I noticed that in no-CORS mode, the request type is opaque, so I can't get the response body (see [this answer](https://stackoverflow.com/questions/54896998/how-to-process-fetch-response-from-an-opaque-type)). I forced no-CORS mode because the hostname `localhost` in the request URL seems to disable CORS (see [this question and answers](https://stackoverflow.com/questions/10883211/why-does-my-http-localhost-cors-origin-not-work) due to the ngnix setting for our CodePod running on `localhost`. A proxy or any other may need to add to our ngnix server to enable CORS.

@lihebi Do you have any idea or suggestion with it?

ui/src/components/Canvas.tsx
```diff
@@ -531,18 +531,35 @@ const CodeNode = memo<Props>(function ({
...
+    const requesOptions = {
+      method: "POST",
+      headers: { "Content-Type": "application/json" },
+      body: JSON.stringify({ code: pod.content }),
+      mode: "no-cors" as RequestMode,
+      accept: "image/png",
+    };
+    const res = await fetch("http://localhost:3099/api/cook", requesOptions);
+    const pic = await res.blob();
+    // navigator.clipboard.write([new ClipboardItem({ "image/png": pic })]);
```
lihebi, line 544: I think this `navigator.clipboard.write` works. I can paste the image to Google Slides:

Screenshot 2022-12-26 at 6 19 08 PM

ui/src/components/Canvas.tsx
```diff
@@ -531,18 +531,35 @@ const CodeNode = memo<Props>(function ({
     }
   }, [data.parent, setPodParent, id]);
 
+  const getPicture = async (pod) => {
+    const requesOptions = {
+      method: "POST",
+      headers: { "Content-Type": "application/json" },
+      body: JSON.stringify({ code: pod.content }),
+      mode: "no-cors" as RequestMode,
```
lihebi, line 539: CORS seems problematic. Let's add an Nginx route.

```shell
# nginx.conf
location /carbon {
        rewrite ^/carbon(.*)$ $1 break;
        proxy_pass http://carbon:3000;
}
```

and in compose.yml, add a container:

```
carbon:
    image: local/carbonara
```

The request URL will be `http://localhost/carbon/api/cook`

PR #185, opened 2022-12-27
refactor zustand store

lihebi:
- Split the store into many small [slices](https://github.com/pmndrs/zustand/blob/main/docs/guides/typescript.md#slices-pattern).
- Move large action functions out so that createXXXStore can be concise.
- Cleanup deprecated functions (e.g., old import/exports related ones, computeNamespace)


li-xin-yi:
We may also need to refactor `Canvas.tsx`, move the definition of `codeNode` and `scopeNode` into `components/nodes`. `Canvas.tsx` is also too heavy, it's painful to code on it. :sob:

lihebi:
Yep, I'm working on refactoring Canvas.

PR #193, opened 2023-01-04
Fix: some miscellaneous & tiny bugs 

li-xin-yi:
1. disable `richNode` editing & `codeNode` copying for guest users
2. fix the bug: a guest user sees the pasted pod increases on its size endlessly. 
3. fix the bug: a pasted pod's position is not up-to-date when adding.
4. extend node types when observing adding nodes in `nodesMap`
5. refactor the code about the privilege management in front-end, hide the detailed role info to simplify the code

More details can be found in comments.

Suggestions:
- Many functions in `canvas.tsx` can be moved into `lib/utils.tsx` like `dbtype2nodetype` to make `canvas.tsx` more clear.
- For a guest user, the runtime should not be loaded from the beginning, it wastes resource.

Other works on the way:

- Figure out a way to prevent selecting node by `Space` or `Enter` (see https://github.com/codepod-io/codepod/issues/166)

### DIFF
--- a/ui/src/components/Canvas.tsx
+++ b/ui/src/components/Canvas.tsx
@@ -31,7 +31,8 @@ import { lowercase, numbers } from "nanoid-dictionary";
 
 import { useStore } from "zustand";
 
-import { RepoContext, RoleType } from "../lib/store";
+import { RepoContext } from "../lib/store";
+import { dbtype2nodetype, nodetype2dbtype } from "../lib/utils";
 import {
   useNodesStateSynced,
   resetSelection,
@@ -72,39 +73,6 @@ function getAbsPos({ node, nodesMap }) {
     return [x + dx, y + dy];
   } else {
     return [x, y];
-  }
-}
-
-/**
- * For historical reason, the state.pod.type and DB schema pod.type are "CODE",
- * "DECK", "WYSIWYG", while the node types in react-flow are "code", "scope",
- * "rich". These two functions document this and handle the conversion.
- * @param dbtype
- * @returns
- */
-function dbtype2nodetype(dbtype: string) {
-  switch (dbtype) {
-    case "CODE":
-      return "code";
-    case "DECK":
-      return "scope";
-    case "WYSIWYG":
-      return "rich";
-    default:
-      throw new Error(`unknown dbtype ${dbtype}`);
-  }
-}
-
-function nodetype2dbtype(nodetype: string) {
-  switch (nodetype) {
-    case "code":
-      return "CODE";
-    case "scope":
-      return "DECK";
-    case "rich":
-      return "WYSIWYG";
-    default:
-      throw new Error(`unknown nodetype ${nodetype}`);
   }
 }
 
@@ -162,7 +130,7 @@ function useCopyPaste(reactFlowWrapper) {
 
   const getPod = useStore(store, (state) => state.getPod);
   const nodesMap = useStore(store, (state) => state.ydoc.getMap<Node>("pods"));
-  const role = useStore(store, (state) => state.role);
+  const isGuest = useStore(store, (state) => state.isGuest());
 
   const reactFlowInstance = useReactFlow();
 
@@ -214,7 +182,8 @@ function useCopyPaste(reactFlowWrapper) {
         ...node,
         width: node.width,
         height: node.height,
-        style: { ...node.style, opacity: 1 },
+        // width must be set here otherwise a guest will see its width keeps increasing
+        style: { ...node.style, opacity: 1, width: node.width! },
         data: {
           level: 0,
           name: node.data?.name,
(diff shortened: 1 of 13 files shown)


li-xin-yi:
Update: Fix the selection issue for guest users. Now they can also multi-select pod like other users. (pre-requirement for multi-pod copy by guests)

ui/src/components/Canvas.tsx
```diff
@@ -162,7 +130,7 @@ function useCopyPaste(reactFlowWrapper) {
 
   const getPod = useStore(store, (state) => state.getPod);
   const nodesMap = useStore(store, (state) => state.ydoc.getMap<Node>("pods"));
-  const role = useStore(store, (state) => state.role);
+  const isGuest = useStore(store, (state) => state.isGuest());
```
lihebi, line 133: Two reasons pull me back from this change.

1. Here, `isGuest` is the result of calling a Zustand action function `state.isGuest()`. I have never seen such usage before, and I doubt that it might be incompatible with zustand's ability to "Renders components only on changes". I.e., `isGuest` might not be able to trigger a re-render if `store.role` changes. Thus, I don't think we want to use this pattern.


2. Also, it adds two more actions (`isGuest` and `isOwner`) in Zustand store (which are effectively global functions). We want to keep Zustand store lean, only those states and **state mutators**.


For reference, below is Zustand store's typical usage pattern: state and state mutators:

```ts
const useFishStore = create((set) => ({
  salmon: 1,
  tuna: 2,
  deleteEverything: () => set({}, true), // clears the entire store, actions included
  deleteTuna: () => set((state) => omit(state, ['tuna']), true),
}))
```

Back to our problem, we could define `isGuest` and `isOwner` as functions, and use `isGuest(role)` here, e.g.,

```ts
// store/index.tsx or some utils.tsx
export const isGuest = (role) => role === RoleType.GUEST

// Code.tsx
import {isGuest} from ...
function Component() {
  const role = useStore(..., state.role)
  ...
  return <Box>{isGuest(role) && <Box>I'm guest</Box>}</Box>
}
```

li-xin-yi: I'm sure it must re-render when role changes because I set all the initial role as `GUEST` in the beginning, and it re-enables editing when the repo loads successfully for non-guest users. It's a function call evaluated in real-time, not a function declaration.

I just feel it may be problematic to expose the explicit role info to other components. We just need to know it abstractly. I have two ideas to do that:
1. Instead of setting `role` in `store`, split it to two fields `isGuest: boolean` and `isOwner: boolean`.
2. Similar to your solution, but hide the actual `role` like:

```ts
// store/index.tsx or some utils.tsx
export const checkGuest = (role) => role === RoleType.GUEST

// Code.tsx
import {isGuest} from ...
function Component() {
  const isGuest = useStore(..., checkGuest(state.role))
  ...
  return <Box>{isGuest && <Box>I'm guest</Box>}</Box>
}
```

lihebi: > 1. Instead of setting `role` in `store`, split it to two fields `isGuest: boolean` and `isOwner: boolean`.

This will leave space for bugs, e.g. we can set both `isGuest` and `isOwner` to `true`.

> 2. Similar to your solution, but hide the actual role like:

Sounds good, let's use this.

li-xin-yi:
Update: I finally removed the `enum` type and change it to `role: "OWNER" | "COLLABORATOR" | "GUEST"` (uppercase due to constant declaration).

There are still three known issues in this PR:

1. Auto-unselection on focusing `codeNode`/`richNode` doesn't work when the user is a guest. It seems to inactivate drag-handler settings of nodes when `nodesDraggable=false`, so no matter you click on any place inside the node, it will be selected immediately.
2. There exists accuracy error when writing a float number into DB or in reverse. The node position coordinates become float numbers after dragging. So when you re-enter the repo after some drag operations, you may see some errors threw by `verifyConsistency`:
image

  A possible solution: stop using `!==` to compare floats (many classic issues can be referenced in [IEEE 754](https://en.wikipedia.org/wiki/IEEE_754)), you can use something like:

```ts
if (Math.abs(node1.position.x-node2.position.x) > Number.EPSILON) {
...
}
```

3. I set `editable=false` in `Remirror` component for guest users, the text cursor cannot be displayed in the node anymore, which means a guest cannot even select some texts. This is a totally different behavior from the [official demo in storybook](https://remirror.vercel.app/?path=/story/editors-controlled--editable). But when you double-click on the text, the floating toolbar shows up, you can still bold, underscore, or highlight all texts. To solve the problem *temporarily*, you can implement the "read-only editor" manually by:
```ts
            onChange={(parameter) => {
              if(isGuest && parameter.tr?.docChanged) return;
              ...
            }}
```
But I feel many features are really wired in `Remirror`. If you have time, can you investigate it more and find any clue?

lihebi:
> 3\. the text cursor cannot be displayed in the node anymore, which means a guest cannot even select some texts.

It is because reactflow nodes have `user-select: "none"` CSS property. We just need to set it back:

```
<Box sx={{userSelect: "text", cursor: "auto"}}> ... </Box>
```

Let's fix it in another PR.

lihebi (approved):
Thank you, Xinyi!

lihebi:
> ```js
> sx={{userSelect: "text", cursor: "auto"}}
> ```

For reference, PR #196 allows text selection in result block.

PR #197, opened 2023-01-04
add devMode

lihebi:
Add a new "dev mode" switch in the sidebar. When it is on, the ID of all code pods will be displayed.

Screenshot 2023-01-04 at 11 00 31 AM


forrestbao:
Should it be called "debug mode"?

lihebi:
Agree, "debug mode" makes more sense. Changed in #206.

PR #199, opened 2023-01-05
Minor fixes

li-xin-yi:
1. Fix the read-only selection / toolbar issues
2. Fix the float comparison issue (see [here](https://github.com/codepod-io/codepod/pull/193#issuecomment-1370637990))
3. Fix the selection problem caused by `Enter`/`Space` in `richNode`


lihebi (approved):
Thanks!

PR #206, opened 2023-01-10
rename Dev Mode to Debug Mode

lihebi:


li-xin-yi:
I think we can also put the toggle into the context menu on canvas, then users don't have to expand the sidebar when they want to operate on it.

PR #207, opened 2023-01-12
Fix: Minor fixes

li-xin-yi:
1. Unselect all pods when a codeNode or richNode is focused.
2. Enlarge the difference when comparing the position value. `Number.EPSILON` is too small.

### DIFF
--- a/ui/src/components/Canvas.tsx
+++ b/ui/src/components/Canvas.tsx
@@ -121,7 +121,9 @@ function verifyConsistency(nodes: Node[], nodesMap: YMap<Node>) {
       );
       return false;
     }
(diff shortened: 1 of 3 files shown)


li-xin-yi:
Update:

Enable debug mode for `richNode`, clean up the hard code hacking of disabling selection on focus for richNode (because it is solved in the latest reactflow by default, we no longer need to listen to keyboard event).

Known issues:

1. I suggest moving toggles in "site settings" into the right-click menu to make the sidebar looks more light.
2. nodes are re-rendered every time their position changes because `xPos` and `yPos` changes, causing a lot of overhead, which is even worse in `richNode` because we don't memo the editor.
image

Let's solve them in the future work if necessary.

lihebi (approved):
Thanks!

PR #214, opened 2023-02-01
Fix: incomplete dashed cutting hint

li-xin-yi:
Fix:

1. Use the original border and change its style as red & dashed when the pod is under cutting

image

2. Fix the issue that resizing a codeNode/richNode will never trigger the store and DB sync.


lihebi (approved):
Thanks!

PR #218, opened 2023-02-23
add python and JSON file export

lihebi:
UI in the sidebar:

Screenshot 2023-02-23 at 5 34 34 PM

When clicked, the files will be downloaded. The exported python file with indentation:

Screenshot 2023-02-23 at 5 37 09 PM

The exported JSON file:

Screenshot 2023-02-23 at 5 35 19 PM

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,7 +3,7 @@
 Codepod provides the interactive coding experience popularized by Jupyter, but
 with scalability and production-readiness. Users can still incrementally build
 up code by trying out a small code snippet each time. But they would not be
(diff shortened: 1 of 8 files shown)


forrestbao:
When there are multiple pods, how are their orders determined in the exported Python files? Can we allow users to customize the order by drawing arrows? Or can allow users to select ordering schemes, like based on execution order?

lihebi:
Yes, we’d handle those in next PR and make the .py runnable. This PR for now is mainly for backing up purpose before we have a proper VCS.

PR #219, opened 2023-02-25
feat: add updatedAt field; order projects by last update time

lihebi:
Now each project (repository) will record the updatedAt and createdAt fields. The project list will be ordered based on last update time, with most recent showing first.

Screenshot 2023-02-25 at 11 58 30 AM


lihebi:
Update:
1. Use "last viewed" instead of "last updated" to sort the projects, the same behavior as Google Docs.
2. In DB schema, the data is stored in a UserRepoData table with one entry for each `<user,repo>` pair, so that each user owns a different setting for each repo.

```prisma
model UserRepoData {
    id         String   @id
    user       User     @relation(fields: [userId], references: [id])
    userId     String
    repo       Repo     @relation(fields: [repoId], references: [id])
    repoId     String
    accessedAt DateTime @default(now()) @updatedAt
    dummyCount Int      @default(0)
}
```

PR #226, opened 2023-02-27
feat: add arrow-import runtime semantic

lihebi:
The import dependency of programs is not a tree, but a DAG (Directed Acyclic Graph). For example, an utility function may be used by many scopes. A tree structure cannot handle this, as there's only one parent for a tree node.

Now we can add an arrow from one scope to another to mark a manual import relation.

Screenshot 2023-02-27 at 5 18 33 PM


forrestbao:
Using arrows occasionally or in short distances is fine. But 
1. the repo may be prone to become go-to. 
2. drawing arrows is a laborious mouse operation, unscalable
3. too many arrows make the canvas messy. 

So besides arrows, we can have use some code-based solution. For example,  enrich the `@export` decorator, like `@export('siblings')` or `@export('recursive')`

Issue #204, opened 2023-01-07
Directed autolayout


forrestbao:
In terms of autolayout with directions (either data dependencies, call dependencies, etc.), i think we can borrow the algorithm behind the Mermaid language. According to my experience, it does not work well for cyclical relations. But if just trees, it works good. It even supports subgraphs, which are like our scopes. 

[One recent example I used](https://github.com/SigmaWe/SmartCommand#architecture)

Issue #176, opened 2022-12-25
auto-shovel pods/scopes that will overlap due to width/height expansion


forrestbao:
I am trying to use CodePod to develop a [real bioinformatics program](https://app.codepod.io/repo/dg0bns4a8yxqpvapb0ni) and I have found a series of pains. This is the first one. 

**Problem**

I spend a lot of time on manually repositioning pods or scopes. This greatly reduces my development speed. 

There are two cases: 
1. If I expand a **scope's width or height**, or **the width of a pod**, manually, pods and scopes in the way do not move away automatically. Instead, they overlap. So I have to first make space for the expansion and then expand. 
2. For a code pod, as the code gets longer (thus **the height of a pod** increases), it overlaps with pods or scopes that were originally below it. Like the case above, I have to manually find spaces again. 

**Expected change**

We do not need fancy auto-layout to solve this problem. Just do auto-shoveling. When the width (height) of a pod/scope expands, automatically push the pods/scopes in the way to the right (down), recursively, so users do not have manually make the space. 

For the scope, I really want the automatic bubbling #151 to be implemented so I can save time from manually adjusting the shape the of a scope.

li-xin-yi:
Will a growing pod/scope push other pods nearby away, or we just let them overlap with other peers?

forrestbao:
> Will a growing pod/scope push other pods nearby away, or we just let them overlap with other peers?

Push them away.

Issue #49, opened 2022-11-13
Call-graph based inner-scope auto-layout


lihebi:
Features
- analyze code snippets to understand:
  - the function name
  - the call graph
- According to the call graph, auto-layout all pods in the selected scope and add arrows.

Something like this:

<img width="622" alt="Screenshot 2022-11-10 at 7 46 51 PM" src="https://user-images.githubusercontent.com/4576201/201503159-0bcb5ae1-9309-4ee1-88bf-27931eed9343.png">

senwang86:
I think this's a default option, but more importantly, Codepod should honer user's manual layout, and not to reset it. My personal opinion is "auto alignment" instead of "auto layout".

forrestbao:
The issue is what if the manual layout has to be broken? Say you want to add a new pod between two pods. Should the two pods be shoveled out to make space for the new pod?  This is where we need a good algorithm instead of simply using force which pushes things out of alignment. Can we add some regularization terms on the force?

senwang86:
> The issue is what if the manual layout has to be broken? Say you want to add a new pod between two pods. Should the two pods be shoveled out to make space for the new pod? This is where we need a good algorithm instead of simply using force which pushes things out of alignment. Can we add some regularization terms on the force?

It's perhaps a result of my personal habits, I am OK with the way that Mac manages desktop icons (see 00:48 of this [YouTube video](https://youtu.be/qIGvcjOTiwo)), and Mac even permits icons to overlap. The phrase "clean up" is what I prefer to use instead of finding a creative way to move the two pods.

PR #231, opened 2023-03-01
[draft] add d3-based auto-layout

lihebi:
This is an ongoing draft of adding d3-based auto-layout. There're still many details to be figured out.

An "auto-layout" button is added to the scope's toolbar. Once clicked, the pods inside the scope shall be auto-layouted. A global auto-layout button shall auto-layout the entire canvas.

Before auto-layout:
Screenshot 2023-03-01 at 12 42 11 PM

After auto-layout:
Screenshot 2023-03-01 at 12 42 30 PM

Known issues:
- The layout position didn't consider the scope's position, thus the layout is not inside the scope.


lihebi:
This PR has three layout algorithms:
- d3-force: this is not ideal, as the layout results are dependent on the original position of the nodes. Our layout engine should be purely functional w.r.t. the edge and parent relations among the pods.
- d3-hierarchy: this does not support nodes with variable width and height.
- d3-flextree: this should be the most promising one.

Details of how to use d3 to auto-layout our pods:
- create a d3 compatible graph data structure from our nodes
- call d3's auto-layout algorithm on the newly created data structure. We need to choose what kind of force to use
- get the updated positions of the d3 data structure and set the corresponding reactflow node positions.

lihebi:
Related issues:
#204
#49 
#176

lihebi:
merged #245

PR #232, opened 2023-04-20
Fix (better drag-n-drop): Scope Handler & Scope right-click Menu for adding new nodes

li-xin-yi:
1. Limit the draggable area of a scope to only the title bar (the same position as code nodes)
2. Right-click Menu in a scope node, you can create new nodes right in this scope instead of creating them outside and then dragging them into

image

Note: To make sure the created node totally inside the scope, it will adjust the position of the new created node like how you drag an existing node into a scope. However, the adjustment is pretty limited because we still lack an auto-layout logic of all nodes. Something funny may happen if you create a scope in a newly-created scope because they have the exact size (default size), you can hardly separate them then. **Refactoring is required after auto-layout/resize is implemented**.

### DIFF
--- a/ui/src/components/Canvas.tsx
+++ b/ui/src/components/Canvas.tsx
@@ -524,10 +524,22 @@ function CanvasImpl() {
   const [showContextMenu, setShowContextMenu] = useState(false);
   const [points, setPoints] = useState({ x: 0, y: 0 });
   const [client, setClient] = useState({ x: 0, y: 0 });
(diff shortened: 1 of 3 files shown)


lihebi (approved):
Works great, thanks!

lihebi:
SG, let's use the auto-hide handle in your solution one first. We kind-of already have something similar to solution 2, the "cut" button. The UI logic could be refined though.

PR #236, opened 2023-04-22
feat: add auto shrink

lihebi:
Now auto-shrink or enlarge the scopes to fit its children. This PR also adds a Command Palette triggered by Ctrl+k (or Cmd+k on macOS) via [kbar](https://github.com/timc1/kbar); auto-layout can be invoked in the Command Palette. In the future, we could add a setting option to run auto layout automatically. Screenshot:

Before auto-layout, and the Command Palette:

Screenshot from 2023-04-21 18-13-01

After auto-layout:

Screenshot from 2023-04-21 18-13-13

### DIFF
--- a/ui/package.json
+++ b/ui/package.json
@@ -28,6 +28,7 @@
     "crypto-js": "^4.1.1",
     "formik": "^2.2.9",
     "graphql": "^16.6.0",
+    "kbar": "^0.1.0-beta.40",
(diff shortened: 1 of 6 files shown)


lihebi:
Also, since we have auto layout now, the following features are removed:

1. restrict pods to be inside a scope (the `extent: parent`)
2. auto-align the node to keep it bound in the scope when moving into the scope.

PR #239, opened 2023-04-22
feat: add run-chain to run a chain of pods

lihebi:
Now there's a "run chain" button to run a chain of pods connected with arrows.

Screenshot from 2023-04-21 18-48-57


lihebi:
Update: wait for the execution result before running the next pod, i.e., run the chained pods one by one. This chained behavior applies to the pods with the vanilla `Run` button as well. Show a loading indicator of all pods scheduled for running. Screenshot:

Screenshot from 2023-04-27 12-00-01

PR #247, opened 2023-04-29
auto hide resize handles for code and rich pods

lihebi:


lihebi:
Now the pods and scopes look clean and minimal unless hovered over.

Screenshot from 2023-04-29 10-03-33

Hovered over:

Screenshot from 2023-04-29 10-03-42

PR #270, opened 2023-05-04
use rectangle collision in d3-force auto layout

lihebi:
Now use rectangle collision instead of circle collision in #245 

Reference: https://observablehq.com/@roblallier/rectangle-collision-force


lihebi:
Now the auto-layout is pretty usable. Pods will only be pushed away when overlapped. The auto-run auto-layout setting is turned on by default.

PR #283, opened 2023-05-06
feat: add export to svg

lihebi:
There's a "download image" button in the sidebar. Once clicked, a high-quality SVG file will be generated (for the **current viewport** of the canvas). The file name is `<reponame>-<repoId>-<ISOdate>.svg` so that it is easy to backup your work with version info.

The SVG file can be big, e.g., 5MB. You can then convert it into pdf, which makes it 150KB.

The UI screenshot:

Screenshot from 2023-05-06 15-39-55

Generated .svg file:

test-download-9f02svi2nk1bpetqeohs-2023-05-06T22_38_17 848Z

### DIFF
--- a/ui/package.json
+++ b/ui/package.json
@@ -26,6 +26,7 @@
     "d3-quadtree": "^3.0.1",
     "formik": "^2.2.9",
     "graphql": "^16.6.0",
+    "html-to-image": "^1.11.11",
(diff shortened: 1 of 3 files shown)


lihebi:
Reference: https://github.com/wbkd/react-flow/discussions/1061

PR #284, opened 2023-05-07
feat: allow pod to access to names in all ancestor scopes

lihebi:
Now a pod can access all names defined in its ancestor scopes. The following canvas shows 4 examples:

1. **[new behavior]** function can be accessed in an inner scope
2. **[new behavior]** global variables can be accessed anywhere
3. exported name can be accessed in its parent scope
4. unexported name cannot be accessed in its parent scope

Screenshot from 2023-05-06 17-27-45

Technical staff: this PR also supports ROOT in Node2Children map, and auto-build it.

### DIFF
--- a/ui/src/lib/store/canvasSlice.tsx
+++ b/ui/src/lib/store/canvasSlice.tsx
@@ -315,7 +315,6 @@ export interface CanvasSlice {
 
   node2children: Map<string, string[]>;
   buildNode2Children: () => void;
-  autoLayout: () => void;
   autoForce: (scopeId: string) => void;
   autoForceGlobal: () => void;
 }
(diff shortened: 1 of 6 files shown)


ui/src/lib/store/canvasSlice.tsx
```diff
@@ -855,6 +855,8 @@ export const createCanvasSlice: StateCreator<MyState, [], [], CanvasSlice> = (
           nodesMap.delete(change.id);
           // remove from store
           get().deletePod(client, { id: change.id });
+          // run auto-layout
+          get().autoForceGlobal();
```
lihebi, line 859: This PR also runs auto-layout on deleting a node.

PR #299, opened 2023-05-10
fix: build node2children map on node add

lihebi:


lihebi:
This fixes a regression introduced by #284 where the `state.node2children` map is used to get pod hierarchy in `runtimeSlice.tsx`

PR #312, opened 2023-05-18
Feat: Fork a Repo

li-xin-yi:
image

Make a copy of the current repo and navigate to the copied repo (untitled) in-place. 

Because the only difference is the repo id (URL), so it looks exactly the same after clicking on the button, just keep an eye on the address in your browser.

### DIFF
--- a/api/src/resolver_repo.ts
+++ b/api/src/resolver_repo.ts
@@ -216,7 +216,7 @@ async function deleteEdge(_, { source, target }, { userId }) {
   return true;
 }
 
-async function createRepo(_, { id, name, isPublic }, { userId }) {
+async function createRepo(_, {}, { userId }) {
   if (!userId) throw Error("Unauthenticated");
   const repo = await prisma.repo.create({
     data: {
@@ -497,7 +497,7 @@ async function deletePods(_, { ids }: { ids: string[] }, { userId }) {
   return true;
 }
 
-async function copyRepo(_, { repoId, name, isPublic }, { userId }) {
+async function copyRepo(_, { repoId }, { userId }) {
   // Find the repo
   const repo = await prisma.repo.findFirst({
     where: {
@@ -506,7 +506,6 @@ async function copyRepo(_, { repoId, name, isPublic }, { userId }) {
     include: {
       pods: {
         include: {
-          children: true,
           parent: true,
         },
       },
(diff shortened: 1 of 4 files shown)


lihebi:
I think we'd call it "make a copy" instead of "fork" because we're not tracking the upstream repo like what GitHub did. This "make-a-copy" feature is good enough, as we don't have a proper VCS to support the real fork now.

Also, it would be nice to name the new repo "copy-of-XXX" instead of leaving it untitled, where XXX could be:
* if the original repo has the title, use that title
* otherwise, use the original repo's ID

li-xin-yi:
> I think we'd call it "make a copy" instead of "fork" because we're not tracking the upstream repo like what GitHub did.

I agree.

Besides, open a new tab or navigate to the copied repo in place, which method do you prefer?

lihebi:
Good idea, I’d prefer opening a new tab.

ui/src/pages/repo.tsx
```diff
@@ -180,6 +193,19 @@ function RepoWrapper({ children, id }) {
               Share
             </Button>
           }
+          forkButton={
+            <Button
+              endIcon={<ContentCopyIcon />}
+              onClick={async () => {
+                const result = await copyRepo();
```
lihebi, line 200: This is interesting; I didn't know we could await for the mutation function. Looks like we indeed can.

api/src/resolver_repo.ts
```diff
@@ -497,6 +497,68 @@ async function deletePods(_, { ids }: { ids: string[] }, { userId }) {
...
+      ...pod,
+      id: idMap.get(pod.id),
+      parent: pod.parent ? { id: idMap.get(pod.parent.id) } : undefined,
+      repoId: id,
+      parentId: pod.parentId ? idMap.get(pod.parentId) : undefined,
+    };
+  });
+
+  // Add all nodes without parent/child relationship to the new repo.
+  // TODO: it updates the parent/child relationship automatically somehow,maybe because the parentId? Try to figure out why, then refactor addPods method.
```
lihebi, line 549: Two comments:
1. The parent/children relationship is a **single** relation (named the `PARENT` relation in the schema). Thus, updating one should automatically update another. We should probably only update the `parent` field and not touch the `children` field at all.
2. Yep, providing the `parentId` field seems to be equivalent of using `{connect: {parent: {id}}}` Prisma clause.

lihebi (approved):

PR #216, opened 2023-02-20
[draft] add file export and github sync

lihebi:
# Introduction
This PR implements file export and GitHub sync. Users can sign in and link their GitHub account on the profile page. Each project can be linked to a specific GitHub repo, and an "export to github" operation will export the project to files and commit to the GitHub repo.

This PR is **only a draft**, there're still some details left, e.g., how to maintain the git repo files when we have multiple replicas of the API servers. We'll **not** use GitHub sync for now.

# Screenshots

login and link GitHub account:

<p float="left">
  
   
</p>

link a project to a GitHub repo:

<p float="left">
  
   
</p>

PR #328, opened 2023-06-09
Feat: AI Code AutoCompletion with Codeium Support

li-xin-yi:
### Usage

Right-click the pane and select this option:
image
Then it prompts a tab and requires your Codeium token obtained from this tab:
image
image

After that, try to type some prompts in code editor, it gives inline completion suggestions:

image

### Tips
1. Turn off the Codeium Chrome extension if installed
2. Run `npx prisma migrate dev` in docker api's shell and then restart

**Don't merge, just for test, many issues may exist**

### Todo
1. clean up
2. handle exceptions
3. better UI and logic implementation for Codeium auth? (how about anonymous users?)
4. fix bugs with storing api keys
5. only use the context in the same editor, we may add more code editor contents in the following PRs.


api/prisma/schema.prisma
```diff
@@ -51,6 +51,8 @@ model User {
     Repo         Repo[]         @relation("OWNER")
     sharedRepos  Repo[]         @relation("COLLABORATOR")
     UserRepoData UserRepoData[]
+
+    codeiumAPIKey String? @default("")
```
lihebi, line 55: I think we'd better not store user tokens in the database, for security and privacy policy purposes.

Let's use the browser cache (`localStorage`) instead? And let's make all the Codeium API requests directly from the user's browser.

I'm having the same concern in #216, and I plan to use `localStorage` for GitHub token as well.

senwang86:
Thanks for the hard work, @li-xin-yi !
A couple of thoughts regarding the To-do list:
> 3. better UI and logic implementation for Codeium auth? (how about anonymous users?)

- If the enabling/disabling behavior is not tied to certain pods/scopes, it may be better to move the menu item from "right click" to "sidebar", making it easier for users to enable/disable Codeium support.
- As @lihebi mentioned, to reduce the UI complexity, we might not require an extra pop-up window for the Codeium token. Once the functionality/connection to Codeium server is stabilized, we can improve the UI work, e.g., by subsequently developing a "User setting" page to add various tokens. 
- By default, we can register another Codeium account (e.g., anony_codepod) for anonymous users.

lihebi:
> By default, we can register another Codeium account (e.g., anony_codepod) for anonymous users.

I think people should bring their tokens; it's their business and pricing plans with Codeium (though it is free). Otherwise, we need to obtain some enterprise licenses with Codeium to provide it for our users.

Anonymous users can still set a token. If we use `localStorage` to store the token, it doesn't matter whether the user is registered or not.

api/prisma/schema.prisma
li-xin-yi: I agree. [`localStorage`](https://developer.mozilla.org/en-US/docs/Web/API/Window/localStorage) seems available across all browsers. Should we use it to store users' API keys or their tokens?

li-xin-yi:
Update:

I also keep the option on the sidebar (BTW, I think many options can also be kept on sidebar, for example, show line numbers)

I'm considering about storing stuffs in the browser's local storage, not only the token, but the option to turn on/off some feats, such as auto-completion, show annotations, or maybe the last viewpoint...

api/prisma/schema.prisma
li-xin-yi: Another strange question: if one stored his/her own token before, after he/she logs out and signs in with another account, should the stored token remains?

lihebi:
> I also keep the option on the sidebar (BTW, I think many options can also be kept on sidebar, for example, show line numbers)

I have a `settingSlice.ts` to keep all the `localStorage` settings. You can put the settings there.

https://github.com/codepod-io/codepod/blob/0471799c49f601b9f1a840b5965e9fe9ae43d8ad/ui/src/lib/store/settingSlice.tsx#L20-L50

api/prisma/schema.prisma
lihebi: In this sense, `localStorage` is not a good option to store either settings or user tokens. We should indeed use the database.

lihebi:
> I'm considering about storing stuffs in the browser's local storage, not only the token, but the option to turn on/off some feats, such as auto-completion, show annotations, or maybe the last viewpoint...

Let's think twice about `localStorage`, it has several downsides:
1. the most important (as you mentioned): the settings/tokens should come and go with the users. In this sense, we should not use `localStorage`.
2. it does not support cross-browser syncing.

I think we could put a few simple things like "debug mode" in `localStorage`, and clear (reset) it after user logouts. All other settings and tokens that are expected to sync cross browsers should be kept in the database.

li-xin-yi:
Update:

image

1. Allow user to bring their own API key (token) by clicking on the small help icon right after the auto-completion option:
image

2. Store the user's API key in DB, the option of turn on/off auto-completion/custom-API in the browser's local storage. 
image

(warning if no stored API key for yourself)
3. Set a default API key if user doesn't specify their own key.
4. Ban the auto-completion for guest users (not anonymous users, refer to the users who doesn't have a write access to the repo)

To add the default Codeium API key: add one line in `compose/dev/.env`:

```
CODEIUM_API_KEY=<your_api_key>
```

To get the plaintext of your Codeium key, verify your key in the dialog above, check the console log, it prints your updated api key (I will remove it for final clean-up, just for tests)

li-xin-yi:
Update: when no default API specified in `.env` file, the "auto-completion" switch is in the off state, when trying to turn it on, the setting window shows up.

ui/src/components/CanvasContextMenu.tsx
```diff
@@ -35,11 +37,10 @@ const ItemStyle = {
...
   if (!store) throw new Error("Missing BearContext.Provider in the tree");
-  const showLineNumbers = useStore(store, (state) => state.showLineNumbers);
-  const flipShowLineNumbers = useStore(
-    store,
-    (state) => state.flipShowLineNumbers
-  );
+  const client = useApolloClient();
+  const autoCompletion = useStore(store, (state) => state.autoCompletion);
+
+  console.log("autoCompletion", autoCompletion);
```
lihebi, line 43: clean this?

ui/src/lib/codeiumClient.ts
```diff
@@ -0,0 +1,126 @@
...
+  ): Promise<GetCompletionsResponse | undefined> {
+    this.abortController?.abort();
+    this.abortController = new AbortController();
+    const signal = this.abortController.signal;
+    try {
+      const getCompletionsPromise = this.client.getCompletions(request, {
+        signal,
+        headers: this.getHeaders(request.metadata?.apiKey),
+      });
+      console.log("codeium request", request);
```
lihebi, line 67: This is printed very often. Change it to `console.debug` so that user's console stays clean.

lihebi: Or we could show an indicator in a "status bar" (or the toolbar near the "auto-complete" switch) to show that we are doing a Codeium request.

ui/src/components/SettingDialog.tsx
```diff
@@ -0,0 +1,174 @@
...
+      setStatus("error");
+      setMessage("Token cannot be empty");
+      setInfoShow(true);
+      return;
+    }
+    try {
+      const { api_key, name } = await registerUser(token);
+      if (api_key === "" || api_key === undefined) {
+        throw new Error("Invalid token");
+      }
```
lihebi, line 68: Let's also provide a way so that people can clear their tokens?

lihebi (commented):
Thanks, Xinyi. The PR looks and works very well. A few minor comments.

ui/src/components/MyMonaco.tsx
lihebi, line 1: I feel these changes are leftovers (to be removed) from your code to test Codeium plugins.

li-xin-yi: Yes, definitely my fault, I forgot to clean them up.

ui/src/components/SettingDialog.tsx
li-xin-yi: Sounds good. Added!

ui/src/lib/codeiumClient.ts
li-xin-yi: I like the idea of indicator. Do we have to implement it in this PR? I can add it later.

lihebi: Sg, let’s add it later.

lihebi:
Merging. Thanks, Xinyi!

li-xin-yi:
Forgot to ask: Do I also need to update the [`compose/prod/compose.yml`](https://github.com/codepod-io/codepod/blob/main/compose/prod/compose.yml)?

forrestbao:
> Forgot to ask: Do I also need to update the [`compose/prod/compose.yml`](https://github.com/codepod-io/codepod/blob/main/compose/prod/compose.yml)?

I would be helpful to leave a placeholder there for Codeium API.

lihebi:
[compose/prod/compose.yml](https://github.com/codepod-io/codepod/blob/main/compose/prod/compose.yml) has been deprecated as it was used for hosting CodePod in a production environment. However, the production setup has been transitioned from Docker to Kubernetes (k8s). You can find the updated configuration at https://github.com/codepod-io/codepod-k8s.

Although the prod configuration is no longer recommended for production, it can still be useful for users (non-developers) who want to launch a "local CodePod" with minimal resource requirements. This configuration serves plain HTML and JavaScript instead of using react start, which requires more than 3GB of RAM. Additionally, there is no need to mount the src/ folder since there is no intention to develop the CodePod source code in this scenario.

However, to avoid confusion, it is advisable to temporarily remove the prod configuration. We should release a "local CodePod" in some form once we have a clear plan in place.

ui/package.json
```diff
@@ -63,6 +70,7 @@
     "zustand": "^4.1.3"
   },
   "scripts": {
+    "generate": "chmod 777 /root && rm -rf ./src/proto && npx buf generate --output ./src/",
```
lihebi, line 73: Why do we need `chmod 777 /root`? The app inside the container is mounted at `/app/src/`.

li-xin-yi: Otherwise, `npx buf generate` will fail because I need to create & read a `/root/.config/buf/config.yaml`. I have no idea how to fix it.

lihebi: I see. I'll take a look later.

PR #334, opened 2023-06-19
revise README

forrestbao:


README.md
```diff
@@ -123,3 +153,60 @@ cd ./api/
...
+    image: nginx:alpine
+    ports:
+      - 80:80
+    volumes:
+      - ./nginx.conf:/etc/nginx/conf.d/default.conf
+```
+
+and replace the default port number 80 to a new port number. For example, you can set the port number to 8080 for all occurences of 80. 
+
+Also, comment out the port section of the `ui` container in `CODEPOD_ROOT/compose/dev/compose.yml` as:
```
lihebi, line 183: Let's mention the "development process" somewhere. Basically, the process is:

1. start the docker stack
2. modify some `.ts` files. Upon saving, the server processes will be restarted automatically.
    - How it works:
       - the source files are mounted into the `ui/` or `api/` container in the development stack.
       - the server process monitor the file changes and restart accordingly
4. see the changed behavior in app by refreshing the webpage.

lihebi: In the future, I plan to remove the "hot-reloading" feature of React (by not binding `ui:3000` to `localhost:3000`) because hot-reloading seems to only update the app partially. A manual refresh is safer and more reliable. Hot reloading is reliable for small apps or style changes, but not a complex app like CodePod. But let's keep it as is for now.

README.md
```diff
@@ -80,37 +88,59 @@ Optional:
...
+> ```bash
+> npx prisma migrate dev
+> ```
 
-- One developer changed [the schema](./api/prisma/schema.prisma). He will run
-  `npx prisma migrate dev --name add_a_new_field`. This will generate a
-  migration, e.g. [this
-  migration](./api/prisma/migrations/20221206194247_add_google_login/migration.sql).
-  The schema change along with this migration need to be checked in to git.
-- Another developer pulls the change, then running the `npx prisma migrate dev` (in the api container's shell) to apply the schema change.
```
lihebi, line 113: This is useful but got removed.

lihebi: added back in #363

PR #339, opened 2023-06-20
trigger auto-layout when auto-layout setting is toggled on

chimmyten:
Toggling on the auto-layout setting in the sidebar should now trigger auto-layout for any overlapping pods or scopes.


lihebi:
Thanks, Timmy.

BTW, when your PR resolves an issue, you can [link it](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue) so that the issue will be closed upon PR merging.

chimmyten:
Got it, I'll do that in the future.

PR #343, opened 2023-06-27
[UI] Jupyter Notebook export in Frontend

senwang86:
To export the Codepod repo as one Jupyter Notebook
- In summary,
   - After discussion, the export functionality is implemented in ui server, rather than API server.
   - Add new button "Jupyter Notebook" in the sidebar
   - Build the Jupyter notebook by parsing **Pod** object, ignoring the "result" and "stdout" fields; append the scope structure at the head of each cell.
   - Export cells in geographical order(top-bottom, left-right), following PR [#240](https://github.com/codepod-io/codepod/pull/240)
   - Convert Remirror editor content to markdown, [ref](https://github.com/CodeItQuick/open-sauced/blob/6043be8e6e8c277714daaa7da318d9a2d294b29f/src/components/NoteFormSocial.js#L119C56-L119C56)
- Test
   - 2 pods have 2 scopes, 1 pod w/o scope
   - The exported notebook is rendered by Colab
- ~Bug to fix [[e5b9d1a](https://github.com/codepod-io/codepod/pull/343/commits/e5b9d1a4ad994d7e4a2667fc43459ccec0419f2a)]~
   - ~pod.content is not able to split by newline character~
- Follow up
   - More types of repos should be tested
   - Should test in other platforms
   - ~Add support for rich-text cell~

exportJupyterNotebook

**Note, the export functionality requires the Amazon S3 bucket setup.**


forrestbao:
Cells in Jupyter notebook are linear whereas Codepod is hierarchical and 2D. In what order do we map 2D, hierarchical pods and scopes to linear cells?

senwang86:
> 

If the pods/scopes are connected by "edges", we can order the pods by the flow of edges; otherwise, pods are not ordered.

forrestbao:
Cells in Jupyter notebook are linear whereas Codepod is hierarchical and 2D. In what order do we map 2D, hierarchical pods and scopes to linear cells?

forrestbao:
> > Cells in Jupyter notebook are linear whereas Codepod is hierarchical and 2D. In what order do we map 2D, hierarchical pods and scopes to linear cells?
> 
> If the pods/scopes are connected by "edges", we can order the pods by the flow of edges; otherwise, pods are not ordered.

Un-ordered is not a good idea as a notebook, in many cases, are designated as executable top-down. 
How about this simple strategy? 
1. Members of a scope should be placed above a pod in exported Jupyter notebook. 
2. Within a scope, including the global scope, the order of pods after exporting is: top-down and then left-right? In your screenshot at the top, `pod_1` should be exported first, then `pod_2` and finally `pod_wo_scope`? 
@lihebi

lihebi:
> 2. top-down and then left-right

Yes, this is what I have in my mind.

> 1. Members of a scope should be placed above a pod

Say we have this:

```
pod 1
scope A
  pod A.1
  pod A.2
pod 2
```

Do you mean that we should put `pod A.1` above `pod 1`? That doesn't make too much sense to me.

Or we could use markdown sections with different levels of titles to indicate the scope. Jupyter has a [ToC](https://jupyterlab.readthedocs.io/en/latest/user/toc.html) section to display the levels. However, when we export rich text nodes to markdown cells, they will mess up with the ToC and levels.

image

ui/src/lib/store/podSlice.tsx
```diff
@@ -91,6 +98,20 @@ export const createPodSlice: StateCreator<MyState, [], [], PodSlice> = (
...
     ),
+  setPodRichContent: ({ id, richContent }) =>
+    set(
+      produce((state) => {
+        let pod = state.pods[id];
+        if (pod.type != "RICH") {
+          return;
+        }
+        pod.richContent = richContent;
+        pod.dirty = true;
```
lihebi, line 109: No need to set `pod.dirty=true` here, because `pod.richContent` doesn't need to be saved to DB. Also, this doesn't change the behavior, as `pod.dirty` is already set to `true` in `setPodContent`.

ui/src/components/Sidebar.tsx
```diff
@@ -534,6 +534,168 @@ function ExportJSON() {
...
+        name: repoName,
+        kernelspec: {
+          name: "python3",
+          display_name: "Python 3",
+        },
+        language_info: { name: "python" },
+        Codepod_version: "v0.0.1",
+      },
+      nbformat: 4,
+      nbformat_minor: 0,
```
lihebi, line 668: The latest nbformat version seems to be [5.9.0](https://nbformat.readthedocs.io/en/latest/changelog.html#id2). Should we use that instead?

lihebi (commented):
Thanks, Sen! Overall looks great. A couple of minor comments.

forrestbao:
> > 1. Members of a scope should be placed above a pod
> 
> Say we have this:
> 
> ```
> pod 1
> scope A
>   pod A.1
>   pod A.2
> pod 2
> ```
> 
> Do you mean that we should put `pod A.1` above `pod 1`? That doesn't make too much sense to me.

Sorry, I made a typo. 

Say we have this : 

```
pod 1
scope A
   pod A.1
   pod A.2
   scope A.B
      pod A.B.1
      pod A.B.2
 pod 2
```
then pod A.B.1 and pod A.B.2 should be above pod A.1 and pod A.2 in the exported Jupyter notebook. Makes sense?

ui/src/components/Sidebar.tsx
senwang86: Colab still uses `nbformat:4` in their exported notebook, and it doesn't support the 5.9.0 yet. 

Screen Shot 2023-06-30 at 9 19 24 PM

ui/src/lib/store/podSlice.tsx
senwang86: Updated in [1ceac12](https://github.com/codepod-io/codepod/pull/343/commits/1ceac12c6675a6238ecc010e12845ee240495e9d)

senwang86:
> > > 1. Members of a scope should be placed above a pod
> > 
> > 
> > Say we have this:
> > ```
> > pod 1
> > scope A
> >   pod A.1
> >   pod A.2
> > pod 2
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Do you mean that we should put `pod A.1` above `pod 1`? That doesn't make too much sense to me.
> 
> Sorry, I made a typo.
> 
> Say we have this :
> 
> ```
> pod 1
> scope A
>    pod A.1
>    pod A.2
>    scope A.B
>       pod A.B.1
>       pod A.B.2
>  pod 2
> ```
> 
> then pod A.B.1 and pod A.B.2 should be above pod A.1 and pod A.2 in the exported Jupyter notebook. Makes sense?


exportJupyterNotebook

Updated in [1ceac12](https://github.com/codepod-io/codepod/pull/343/commits/1ceac12c6675a6238ecc010e12845ee240495e9d).

lihebi (approved):
Cool, thanks! Merging.

Issue #365, opened 2023-07-11
`Unknown error` with Codeium


forrestbao:
I am trying to bring-your-own-token. But I got a pop up `unknown error`. 

![unknown_error](https://github.com/codepod-io/codepod/assets/438579/d225c58f-72c2-46b1-899f-8ba6a545cc7a)

lihebi:
I got this error as well because I was using an existing token. You probably need to click "get token" to get a new token. The Codeium token seems to be valid for one-time use.

forrestbao:
Then the error message is better to be "please get a new Codeium token". It's more specific than "Unknown error".

lihebi:
Yes, we should have a better message. Did it solve your problem?

forrestbao:
Yes, getting a new token solved my problem.

li-xin-yi:
Sorry for missing this issue due to a long medical care in clinic these days.

I agree "Unknown Error" is too ambiguous, and I need to specify the error message. But I need your help to determine the error type among invalid, expired, and other wrong token. I tried to update my token several times, but I can't reproduce this error.

Can anyone provide me one expired token via DM on telegram?

Thanks.

lihebi:
I think I used my token (and never used your api key), but got the same error. But I don’t know if I can reproduce the error. As long as we have a message to remind the user to try a new token upon this error, we should be fine.

PR #370, opened 2023-07-13
[Fix] Specific error message for Codeium verification failure 

li-xin-yi:
- Add more details (http response status code) when it fails to update the given token.

close https://github.com/codepod-io/codepod/issues/365


lihebi:
Thanks!

PR #374, opened 2023-07-14
[UI] Allow importing the Python script

senwang86:
## Summary 
- Allow importing a Python script to a pod.
- Change right-click context menu item
- The imported Python script is put into a scope rather than "ROOT". 

## Test

importLocalCode

### DIFF
--- a/ui/src/components/Canvas.tsx
+++ b/ui/src/components/Canvas.tsx
@@ -686,7 +686,7 @@ function CanvasImpl() {
   const autoLayoutROOT = useStore(store, (state) => state.autoLayoutROOT);
 
   const addNode = useStore(store, (state) => state.addNode);
-  const importIpynb = useStore(store, (state) => state.importIpynb);
+  const importLocalCode = useStore(store, (state) => state.importLocalCode);
   const reactFlowInstance = useReactFlow();
 
   const project = useCallback(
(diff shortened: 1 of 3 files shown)


lihebi:
Thanks! Merging.

Two minor concerns that we might want to address in the future:
1. Now we have one menu, "upload" to support both `.ipynb` and `.py` files. We could have two menus, "upload .ipynb file" and "upload .py file" to make it explicit to users.
2. A Python file could be imported into one pod without the scope. But there's a good chance that we provide an option to automatically separate a .py file into multiple pods, e.g., one pod per function. In that case, the scope is needed.

senwang86:
> Thanks! Merging.
> 
> Two minor concerns that we might want to address in the future:
> 
> 1. Now we have one menu, "upload" to support both `.ipynb` and `.py` files. We could have two menus, "upload .ipynb file" and "upload .py file" to make it explicit to users.

I initially added a new item for `.py,` but then realized that, to Codepod, there might not be much of a difference between `.ipynb` and `.py`, then I merged them together. We can definitely alter the menu in future though.  

> 2. A Python file could be imported into one pod without the scope. 

I also considered whether a pod should be imported into "ROOT"; just for myself, I try to keep the number of pods w/o scopes to a minimum; personally, I feel it's more organized.

> But there's a good chance that we provide an option to automatically separate a .py file into multiple pods, e.g., one pod per function. In that case, the scope is needed.

SG.

PR #345, opened 2023-06-28
contextual zoom initial font size reduction

chimmyten:
When a pod is initially collapsed, the font size for the contextual zoom summary will be smaller. When the canvas is zoomed out past a certain point, the font size of the summary will increase to the previous set size. Fixes #316. 

https://github.com/codepod-io/codepod/assets/105993789/5120bd62-19d6-4384-a903-31a1b8944214

Issue #316, opened 2023-05-25
Font size for summaries in contextual zooming 


forrestbao:
The font size for summaries in contextual zoom maybe initially too big. See the video below. 
The initial font when pods collapsed is even bigger than the code/comment in the pods right before the collapse. In the video, when zoomed out, the font size jumped up when the pods collapsed. 

https://github.com/codepod-io/codepod/assets/438579/29cd1175-ecad-4f0a-b2d8-63da5593ffb0

PR #376, opened 2023-07-16
contextual zoom font size adjustment

chimmyten:
When contextual zoom is active, users can adjust the font size using a slider/input box in the sidebar. Fixes #345. Please let me know any feedback. 

https://github.com/codepod-io/codepod/assets/105993789/968be65b-9d3e-49af-a37e-51685daab866


lihebi:
Very cool!

You used one slider to control the font-size of all scope levels. We actually want to have different fontSize for different levels. The `fontSize` variable computes the different font sizes:

https://github.com/codepod-io/codepod/blob/28c442ee81319909f879c80320eb3f9ec6a48b79/ui/src/components/nodes/Code.tsx#L581

So, it's better to use separate sliders to control each scope level individually. I.e., add six sliders to control and set the six font sizes `56, 48, 32, 24, 16, 8` here.

https://github.com/codepod-io/codepod/blob/28c442ee81319909f879c80320eb3f9ec6a48b79/ui/src/lib/store/settingSlice.tsx#L83-L110

forrestbao:
But are separate sliders for different levels too much? Would letting users set the minimal font be better?

lihebi:
At this stage, we want to tune the parameters (ourselves). Making all font sizes configurable can better suit this need. The sliders won't show when "contextual zoom" is not enabled, so it won't be that visible.

forrestbao:
Good point. @lihebi

forrestbao:
I like input boxes, as long as I can quickly use arrow keys on my keyboard to increase or decrease the value. keystrokes are more effective than mouse drags.

lihebi:
I also thought about using only input boxes last night to save space. But i also liked the sliders for quick operation and more explicit visual appearance. 

Let’s have both the sliders and input boxes in this PR? This setting is at an alpha stage, so we can have some redundant UI here. We can refine the UI once it is ready to release.

chimmyten:
Sounds good 👍

li-xin-yi:
Maybe we can consider hotkeys for those operations at the same time. For example, `ctrl/command` + `shift` + `+/-`.

chimmyten:
Hello, I'm a bit confused about the first property of the contextualZoomParams. https://github.com/chimmyten/codepod/blob/main/ui/src/lib/store/settingSlice.tsx#L85C3-L93C5
On which level would a pod have the "prev" font size? Base-level pods, i.e. pods not placed in any scopes seem to adhere to the "0" font size.

lihebi:
You are right; the `.prev` seems not used/needed. You can ignore it or remove it entirely.

Essentially we just need levels 0,1,2,3 and a default value for all levels beyond 3.

chimmyten:
Got it, thank you.

chimmyten:
I added different sliders to control font sizes for different levels. 

https://github.com/codepod-io/codepod/assets/105993789/fd3c8bf3-dfd0-4ea3-9dba-24ac20437658

A couple of notes/questions:
I ran into this strange behavior where the font sizes of the pods wouldn't actually update with the sliders unless I declared this line of code: https://github.com/chimmyten/codepod/blob/817eb1ff0841b8a14b053997c9ba8618768fb5d2/ui/src/components/nodes/Code.tsx#L565-L566. I couldn't figure out why it was affecting the behavior because I don't actually use this line of code anywhere else. If anyone can explain why that would be great.

I commented out a couple lines of code to prevent the pods from disappearing when the font sizes got too small. https://github.com/chimmyten/codepod/blob/817eb1ff0841b8a14b053997c9ba8618768fb5d2/ui/src/components/nodes/Code.tsx#L587-L594.

Please let me know if there are any issues.

lihebi:
> I ran into this strange behavior where the font sizes of the pods wouldn't actually update with the sliders unless I declared this line of code: https://github.com/chimmyten/codepod/blob/817eb1ff0841b8a14b053997c9ba8618768fb5d2/ui/src/components/nodes/Code.tsx#L565-L566. I couldn't figure out why it was affecting the behavior because I don't actually use this line of code anywhere else. If anyone can explain why that would be great.

The reason is that when your slider changes `store.contextualZoomParams`, the Code/Rich/Scope component won't re-render. 

Your `contextualZoomParams = useStore(...)` line did the trick because it essentially tells React to re-render the component when `contextualZoomParams` changes. But this is error-prone because people won't easily see why this line is needed and thus might remove it accidentally.

The preferred way is to remove `store.level2fontsize` and make the usage of `contextualZoomParams` explicit:

```ts
// utils.ts:
export function level2fontsize(level, contextualZoomParams) {...} // move level2fontsize out of Zustand store

// Code.tsx/Rich.tsx/Scope.tsx:
import {level2fontsize} from "./utils"
function Code() {
  ...
  contextualZoomParams = useStore(...)
  ...
  const fontSize = level2fontsize(node?.data.level, contextualZoomParams);
}

```

chimmyten:
Thank you, that makes sense. I'll make the changes.

lihebi:
> I commented out a couple lines of code to prevent the pods from disappearing when the font sizes got too small. https://github.com/chimmyten/codepod/blob/817eb1ff0841b8a14b053997c9ba8618768fb5d2/ui/src/components/nodes/Code.tsx#L587-L594.

Good catch. These are indeed not needed, because a collapsed scope won't have their children rendered at all.

lihebi:
I applied a little clean up and code formatting. Thanks, Timmy!

chimmyten:
I forgot to link the issue, but this PR should fix #316.

PR #378, opened 2023-07-19
[draft] replace CRA with Vite

lihebi:
This is only a demo/draft for reference purposes.

Create-react-app ([CRA](https://create-react-app.dev/)) isn't recommended by React team anymore [[ref](https://react.dev/learn/start-a-new-react-project#production-grade-react-frameworks)]. Also, we needed a ["react-app-rewired"](https://github.com/timarney/react-app-rewired) walkaround to make Monaco editor in CRA's webpack bundler, which isn't nice. The options are:
* [Nextjs](https://nextjs.org/): it seems too complex.
* [Vite](https://vitejs.dev/): it seems good and fast.

Status: The app is working using Vite. But the unit test throws an exception when initializing the parser (`const parser = new Parser()`).


lihebi:
addressed in #468

PR #382, opened 2023-07-20
[Feat] Add LaTeX syntax support for math in notes

li-xin-yi:
Use [`prosemirror-math`](https://github.com/benrbray/prosemirror-math) to add math support for text editors. Support both math in-line and display.

math-demo

Usage:
- Copy formula directly from some website (e.g., Wikipedia) and paste to a note node.
- To start an in-line math formula, type `$` and give a string of LaTeX code, type a `$` to make a end, then the formula appears.
- To start a displayed math, type `$$` in the beginning of the line and press a space, then you will enter the formula mode, just type LaTeX code.
- To edit an existing math stuff, move your cursor into it (either by key or mouse), the LaTeX code is exposed and you can edit now. To exit the edit mode, just move your cursor out of the math environment.

Known bugs:
1. Displayed math (`$$...$$`) can't be started at the first line in an empty note node


ui/package.json
```diff
@@ -35,6 +36,7 @@
     "graphql": "^16.6.0",
     "html-to-image": "^1.11.11",
     "jwt-decode": "^3.1.2",
+    "katex": "0.13.0",
```
lihebi, line 39: Cool trick 😄

li-xin-yi:
Update:

> Displayed math ($$...$$) can't be started at the first line in an empty note node

I found this bug disappears once I turn off the Grammarly Chrome extension. I can't explain why, but anyway it can be one solution...

lihebi:
> turn off the Grammarly Chrome extension

It works for me even if I have the Grammarly extension on. Maybe this bug doesn't exist.

lihebi:
I love this feature. Thanks, Xinyi!

Issue #88, opened 2022-11-24
Command mode and edit mode


forrestbao:
Jupyter/VIM has two modes, Edit/Insert mode and Command mode (`ESC`+`:`). 
We should support both too? 
In command mode, we can change the interpreter (Python, Markdown, Shell) and in edit mode we edit the code.

lihebi:
Yes, we should have this. Further, when a pod is not in edit mode, we should be able to use anywhere on the pod as the handle. @nibilin33 maybe you can help on this?

As for changing interpreter, let’s do a toolbar button instead. I’ll handle this. We can have the VIM keyboard binding later in another issue/PR.

nibilin33:
what...add a button to change the interpreter in the toolbar?

nibilin33:
what...add a button to change the interpreter in the toolbar?

lihebi:
Yeah, after a second thought, we don’t need to change the interpreter for a pod. We can just create different types of pods at the beginning (Python, shell, markdown, other languages).

forrestbao:
Where do you want to let users select the types of pods? @lihebi

lihebi:
In the right click context menu? In Jupyter, users also choose to create markdown or code cells at the beginning.

forrestbao:
In Jupyter or VSCode's Jupyter, you can edit the type of a pod, code to markdown or back. 

I am against using right clicks. It's not explicit. Maybe just an icon in the toolbar of a pod? Associate a shortcut key with it.  (It's hard to do on laptop touchpads. It's not touch-friendly. And it requires users to keep mind that there are two kinds of clicks. ) 

One thing we need to keep in mind in advance is that we will support mixed runtimes/kernels in one repo. So it's more complex than Jupyter's case: code of only one language and Markdown.

lihebi:
The toolbar sounds good. For touch screen, the user could long press the pod, and pop up some bigger buttons. E.g., this is tg's UI:

<img src="https://user-images.githubusercontent.com/4576201/204877361-fe634905-bcaa-4f8b-a2b6-0ffb328e0264.jpeg" width="300">

PR #398, opened 2023-07-25
[UI][1/n] Edit/Command mode

senwang86:
## Summary
Adding edit and command mode into Codepod regarding issue #88 
- Hitting "Enter" on a Scope will go to its upper-left most child
- Hitting "Enter" on a Code/Rich pod will go to "Edit" mode.
- Hitting "Esc" in "Edit" mode will go to "Command" mode.
- Hitting "Esc" on a pod in "Command" mode will go to its parent

## Test

edit_command_mode

### DIFF
--- a/ui/src/components/nodes/Code.tsx
+++ b/ui/src/components/nodes/Code.tsx
@@ -508,6 +508,8 @@ export const CodeNode = memo<NodeProps>(function ({
   const autoRunLayout = useStore(store, (state) => state.autoRunLayout);
 
   const prevLayout = useRef(layout);
+  const ref = useRef<HTMLDivElement>(null);
+
   useEffect(() => {
     if (autoRunLayout) {
       // Run auto-layout when the output box layout changes.
@@ -658,6 +660,13 @@ export const CodeNode = memo<NodeProps>(function ({
       </Box>
     );
 
+  const keyDownHandler = (event) => {
+    if (event.key === "Escape") {
+      ref?.current?.blur();
+    } else if (event.key === "Enter") {
+      ref?.current?.focus();
+    }
+  };
   return (
     <>
       <Box
@@ -814,6 +823,7 @@ export const CodeNode = memo<NodeProps>(function ({
                 height: "90%",
                 py: 1,
               }}
+              onKeyDown={(e) => keyDownHandler(e)}
             >
               <MyMonaco id={id} fontSize={fontSize} />
 
--- a/ui/src/components/nodes/Code.tsx
+++ b/ui/src/components/nodes/Code.tsx
@@ -508,8 +508,6 @@ export const CodeNode = memo<NodeProps>(function ({
   const autoRunLayout = useStore(store, (state) => state.autoRunLayout);
 
   const prevLayout = useRef(layout);
-  const ref = useRef<HTMLDivElement>(null);
-
   useEffect(() => {
     if (autoRunLayout) {
       // Run auto-layout when the output box layout changes.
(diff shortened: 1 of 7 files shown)


lihebi:
Thanks! `Esc` works in Code cells but seems not in Rich cells.

senwang86:
> Thanks! `Esc` works in Code cells but seems not in Rich cells.

Just checked, it seems that `onBlur()` is correctly fired after pressing `Esc`, while the editor box is failed to to be selected as the code editor.

lihebi:
Still one bug. In the edit mode of a Rich pod, Hitting Esc should select the Rich pod. But now it drops me to its parent scope.

lihebi:
> Still one bug. In the edit mode of a Rich pod, Hitting Esc should select the Rich pod. But now it drops me to its parent scope.

To be more specific, this bug happens when the Rich pod is inside a scope.

The cause is:
- Remirror doesn't stop the propagation of the key-press event; therefore, Esc was executed twice.
- Monaco did stop the event propagation, and thus it works fine in Code pods.

I tried a few methods, but none of them worked yet.

ui/src/components/nodes/Code.tsx
```diff
@@ -508,6 +508,8 @@ export const CodeNode = memo<NodeProps>(function ({
   const autoRunLayout = useStore(store, (state) => state.autoRunLayout);
 
   const prevLayout = useRef(layout);
+  const ref = useRef<HTMLDivElement>(null);
```
lihebi, line 511: Not needed?

lihebi:
Thanks!

forrestbao:
Thanks @senwang86 but sorry I probably come across some corner cases: 

1. When I press "Esc" on a pod, I do not always enter the command mode of its parent. If I first select and enter the parent scope, then I can. But the Escape should work without me doing that first. 
    https://github.com/codepod-io/codepod/assets/438579/3bc802ab-b3c6-4b0b-aa1f-d2d8ea31b9a4
2. When I hit "Enter" on a scope, I sometimes got focused on the child of another scope instead of the scope. 
    https://github.com/codepod-io/codepod/assets/438579/6b8c4f70-1893-4726-8b1d-20f815f0e061
3. When I hit "Enter" on a pod, I do not always go into the Edit mode. 
    https://github.com/codepod-io/codepod/assets/438579/582eb194-041f-49f2-85b1-0fa0e395f080

senwang86:
> Thanks @senwang86 but sorry I probably come across some corner cases:
> 
> 1. When I press "Esc" on a pod, I do not always enter the command mode of its parent. If I first select and enter the parent scope, then I can. But the Escape should work without me doing that first.
>    https://github.com/codepod-io/codepod/assets/438579/3bc802ab-b3c6-4b0b-aa1f-d2d8ea31b9a4
> 2. When I hit "Enter" on a scope, I sometimes got focused on the child of another scope instead of the scope.
>    https://github.com/codepod-io/codepod/assets/438579/6b8c4f70-1893-4726-8b1d-20f815f0e061
> 3. When I hit "Enter" on a pod, I do not always go into the Edit mode.
>    https://github.com/codepod-io/codepod/assets/438579/582eb194-041f-49f2-85b1-0fa0e395f080
> 
> You can test in this repo: https://app.codepod.io/repo/eyusr62dyglelzg7m4jz

Really appreciate the test, I can reproduce the bug, the states are not updated in response to the "drag handle click" event. Let me send a patch later.

lihebi:
BTW, I think Esc is not a good key for jump-to-parent. We have many context menus and pop-up menus, and we need to use `<Esc>` to close those menus. How about `<shift>+<up>` and `<shift>+<down>` for up and down jumps?

lihebi:
> How about `<shift>+<up>` and `<shift>+<down>` for up and down jumps?

Moving the shortcut discussion to #418

PR #399, opened 2023-07-27
[UI] Support exporting pod stdout and result

senwang86:
## Summary
- When exporting, the pod's stdout and result are written to a Jupyter notebook
- "Raw JSON" export functionality is moved to the frontend
- The Jupyter notebook result is imported in addition to the entire notebook
- Bug fix: "execution_count" is returned from runtime server now

## Test

export_pod_result

import_ipynb_result

### DIFF
--- a/ui/src/components/Canvas.tsx
+++ b/ui/src/components/Canvas.tsx
@@ -794,6 +794,8 @@ function CanvasImpl() {
           cellList = JSON.parse(String(fileContent)).cells.map((cell) => ({
             cellType: cell.cell_type,
             cellSource: cell.source.join(""),
+            cellOutputs: cell.outputs || [],
+            execution_count: cell.execution_count || 0,
           }));
           importScopeName = fileName.substring(0, fileName.length - 6);
           break;
--- a/ui/src/components/Sidebar.tsx
+++ b/ui/src/components/Sidebar.tsx
@@ -868,7 +868,10 @@ function ExportJSON() {
         "href",
         "data:text/plain;charset=utf-8," + encodeURIComponent(data.exportJSON)
       );
-      element.setAttribute("download", `${repoName || ""}.json`);
+      const filename = `${
+        repoName || "Untitled"
+      }-${new Date().toISOString()}.json`;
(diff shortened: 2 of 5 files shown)


ui/src/components/Sidebar.tsx
```diff
@@ -860,7 +863,17 @@ function ExportJSON() {
   );
   useEffect(() => {
     if (data?.exportJSON) {
-      download(data.exportJSON);
+      let element = document.createElement("a");
+      element.setAttribute(
+        "href",
+        "data:text/plain;charset=utf-8," + encodeURIComponent(data.exportJSON)
+      );
+      element.setAttribute("download", `${repoName || ""}.json`);
```
lihebi, line 871: Let's add date to the exported filename and keep consistent with the .ipynb export.

https://github.com/codepod-io/codepod/blob/554cdac415a9ce579aa6a0b81a8390b7927a91fe/ui/src/components/Sidebar.tsx#L891-L893

ui/src/components/Canvas.tsx
```diff
@@ -794,6 +794,8 @@ function CanvasImpl() {
           cellList = JSON.parse(String(fileContent)).cells.map((cell) => ({
             cellType: cell.cell_type,
             cellSource: cell.source.join(""),
+            cellOutputs: cell.outputs || [],
+            execution_count: cell.execution_count || 0,
```
lihebi, line 798: We should display the execution count in the output area of a Code pod. This PR already gets packed, let's implement it in another PR.

ui/src/components/Sidebar.tsx
lihebi: Nevermind, I made this minor change. Mergin.

lihebi:
Thanks, Sen!

ui/src/components/Canvas.tsx
forrestbao: `execution_count` is a property of both a code cell and its output (if the output type is `execute_result`) https://github.com/codepod-io/codepod/issues/411

PR #401, opened 2023-07-29
[BE] remove cut-n-paste support

senwang86:
## Summary
The cut-n-paste was introduced in PR #165 , as the moveIntoScope #280 function was implemented, cut-n-paste seems unnecessary, this is a Better Engineering effort to reduce the tech debt. 

- The cut button is removed from the toolbar in Code/Scope pod.

### DIFF
--- a/ui/src/components/Canvas.tsx
+++ b/ui/src/components/Canvas.tsx
@@ -490,7 +490,6 @@ function usePaste(reactFlowWrapper) {
   const pasteEnd = useStore(store, (state) => state.pasteEnd);
   const cancelPaste = useStore(store, (state) => state.cancelPaste);
   const isPasting = useStore(store, (state) => state.isPasting);
-  const isCutting = useStore(store, (state) => state.isCutting);
   const isGuest = useStore(store, (state) => state.role === "GUEST");
   const isPaneFocused = useStore(store, (state) => state.isPaneFocused);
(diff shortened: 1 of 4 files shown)


lihebi:
We were talking about that cut isn't necessary anymore. But after a second thought, I feel that cutting may still be useful in a few use cases:
1. Cut once and paste many times
2. The cutting behavior is also slightly different from moving.
    - **Cutting-n-pasting** doesn't alter anything until clicked
    - **Moving** constantly changes the position of the pod being moved. Thus, DB and other peers see the changes in real-time.

senwang86:
> We were talking about that cutting isn't necessary anymore. But after a second thought, I feel that cutting may still be useful in a few cases:
> 
> 1. Cut once and paste many times
> 2. The cutting behavior is also slightly different from moving.
>    
>    * **Cutting-n-pasting** doesn't alter anything until clicked
>    * **Moving** constantly changes the position of the pod being moved. Thus, DB and other peers see the changes in real-time.

The moving effect of the cutting pod is very fancy and I like it, as it shows that "cutting" is taking place; I think copy-n-paste should apply the same effect to indicate that "copy" is happening. 

I am OK with either choice, while I don't think cutting has a strong use case; If cutting isn't regularly used, we might think about moving the button from the toolbar to another location.

senwang86:
PR #452 address the topic.

Issue #420, opened 2023-08-03
Incorrect pod selection with "Drag Handler" clicking


senwang86:
PR #398 add the Edit/CMD mode, however, clicking the "drag handler" seems set the pod selection incorrectly. To reproduce the bug, 
- Pick a scope `A` and click the "drag handler", right now, the scope `A` is selected
- Use arrow keys to jump across pods
- At any pod/scope, press "Enter", instead of entering the pod for edit mode or jump to the 1st child of the scope, you will go to scope `A`'s first child.

senwang86:
PR #431 fixed this issue.

Issue #418, opened 2023-08-03
Keyboard shortcuts for jumping around


lihebi:
Old:
- <Esc>: jump to parent
- <Enter>: jump to top-left child

Problem: `<Esc>` needs to be reserved for close pop-up menus, e.g., right-click menus.

Proposed:

- <shift>-<up>: jump to parent
- <shift>-<down>: jump to top-left child
- <shift>-<left>: follow indegree edge
- <shift>-<right>: follow outdegree edge

PR #431, opened 2023-08-04
[UI] Edit cmd mode patch

senwang86:
## Summary
PR #398 introduced CMD mode, and issue #418 and #420 reveal some improvement and bugs.
- For #418,
    - To go into a scope, the key combination is adjusted to `SHIFT` + `ArrowDown`
    - To go to a parent scope, the key combination is adjusted to `SHIFT` + `ArrowUp`
- For #420, the reason is that `useJump()` uses `selectedPods` to choose the `from` node, while in a multi-selection scenario, `selectedPods` contains more than 1 node. As a workaround, a new state variable `cursorNode` is added for a single node selection.
   - The toolbar pops-up while jumping to a node
- Misc
    - On a code pod, pressing `SHIFT`+`Enter` will run the pod
    - On a scope, pressing `SHIFT`+`Enter` will run the scope

## Test

CMD_mode_patch

### DIFF
--- a/ui/src/components/Canvas.tsx
+++ b/ui/src/components/Canvas.tsx
@@ -50,6 +50,7 @@ import FloatingEdge from "./nodes/FloatingEdge";
 import CustomConnectionLine from "./nodes/CustomConnectionLine";
 import HelperLines from "./HelperLines";
 import { getAbsPos, newNodeShapeConfig } from "../lib/store/canvasSlice";
+import { htmlToProsemirrorNode } from "remirror";
 
 const nodeTypes = { SCOPE: ScopeNode, CODE: CodeNode, RICH: RichNode };
 const edgeTypes = {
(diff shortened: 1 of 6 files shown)


lihebi:
Thanks, this works well to fix #420.

Do you plan to merge `selectedPods` and `cursorNode` in the next PR? I think jumping around should change the selected nodes.

senwang86:
> Thanks, this works well to fix #420.
> 
> Do you plan to merge `selectedPods` and `cursorNode` in the next PR? I think jumping around should change the selected nodes.

Yes, I plan to have a following PR. I think it's cleaner to use different states for "single-selection" and "multi-selection"; and "multi-selection" could be achieved through keyboard shortcut, say pressing `SHIFT`+`ArrowUP|Down|Left|Right`. 

An array-based `selectedPods` can also handle the "single-click" well (by always taking `selectedPods[-1]`), without bringing in a new state variable.

lihebi:
SG, thanks!

PR #432, opened 2023-08-04
fix: remove anonymous guest user

lihebi:
#242 introduced an anonymous guest user feature to allow users to try out CodePod without signing in.

However, the guest token expires in 30 days. After the token expired, there's no way to log in again. This isn't a good design. After the token expires, the app is stuck at the dashboard loading. The Anonymous feature should be done more properly.

This PR temporarily removes this feature.


lihebi:
Note: DB schema changed.

PR #434, opened 2023-08-04
Tool bar size adjustment when contextually zooming out

### LINKED ISSUE #317: Toolbars shrunk after contextually zoomed out
After contextually zooming out, the toolbar icons or maybe the toolbar itself also gets shrunk, making the toolbar icons too small to be recognized or clicked. See the screenshot below: 

Screenshot 2023-05-25 02:07:42

**Expected change:**
Toolbar icons should keep the size. Only pod/scope **contents** should be resized in contextual zooming.

chimmyten:
The toolbar icons will now increase in size to compensate when zooming out to help with visibility. Fixes #317. 

https://github.com/codepod-io/codepod/assets/105993789/4d4f570d-2d28-47c4-a560-f4b96f8a8c46

I also adjusted the toolbar placement for scopes slightly so it would not block the scope title when zoomed out.

### DIFF
--- a/ui/src/components/nodes/Code.tsx
+++ b/ui/src/components/nodes/Code.tsx
@@ -372,6 +372,9 @@ function MyFloatingToolbar({ id, layout, setLayout }) {
     [clonePod, id]
   );
 
+  const zoomLevel = useReactFlowStore((s) => s.transform[2]);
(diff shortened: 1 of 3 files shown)


lihebi:
Nice, thanks!

PR #436, opened 2023-08-05
[Backend] Refactor ipynb kernel messages serialization

senwang86:
## Summary
Currently, the kernel messages are concatenated and stored in a single object, i.e., `pod.result`. This concatenation behavior creates a few discrepancy with Colab regarding the output results, e.g., it can't produce multiple plots in a single pod, the order of line execution might confuse users (see screenshots in Test section)

### Before
- Screenshot 2023-08-04 at 10 41 48 PM
- Screenshot 2023-08-04 at 10 40 19 PM

### After
Screenshot 2023-08-04 at 10 41 10 PM

- Export/Import also verified

## Follow-up
- The `ResultBlock` in `Code.tsx` needs more tuning
- We might need to support more message types, [Messaging in Jupyter](https://jupyter-client.readthedocs.io/en/latest/messaging.html)

### DIFF
--- a/runtime/src/zmq-utils.ts
+++ b/runtime/src/zmq-utils.ts
@@ -281,7 +281,7 @@ export function handleIOPub_status({ msgs, socket, lang }) {
   );
 }
 
-export function handleIOPub_display_data({ msgs, socket, exec_count }) {
+export function handleIOPub_display_data({ msgs, socket }) {
   console.log("emitting display data ..");
   let [podId, name] = msgs.parent_header.msg_id.split("#");
   let payload = {
@@ -294,13 +294,11 @@ export function handleIOPub_display_data({ msgs, socket, exec_count }) {
     //   transient: ...
     // }
     content: msgs.content,
-    // There's no exec_count in display_data, thus we pass in the session exec_count
-    count: exec_count,
   };
   socket.send(JSON.stringify({ type: "display_data", payload }));
 }
 
-export function handleIOPub_execute_result({ msgs, socket, exec_count }) {
+export function handleIOPub_execute_result({ msgs, socket }) {
   console.log("emitting execute_result ..");
   let [podId, name] = msgs.parent_header.msg_id.split("#");
   let payload = {
@@ -309,7 +307,7 @@ export function handleIOPub_execute_result({ msgs, socket, exec_count }) {
     // result: msgs.content.data["text/plain"],
     // This might contain text/plain, or text/html that contains image
     content: msgs.content,
-    count: exec_count,
+    count: msgs.content.execution_count,
   };
   if (name) {
     console.log("emitting IO result");
@@ -356,7 +354,7 @@ export function handleIOPub_error({ msgs, socket }) {
   }
 }
 
-export function handleIOPub_stream({ msgs, socket, exec_count }) {
+export function handleIOPub_stream({ msgs, socket }) {
   if (!msgs.parent_header.msg_id) {
     console.log("No msg_id, skipped");
     console.log(msgs.parent_header);
@@ -376,7 +374,6 @@ export function handleIOPub_stream({ msgs, socket, exec_count }) {
           // name: stdout or stderr
           // text:
           content: msgs.content,
-          count: exec_count,
         },
       })
     );
@@ -389,7 +386,6 @@ export function handleIOPub_stream({ msgs, socket, exec_count }) {
           payload: {
             podId,
             content: msgs.content,
-            count: exec_count,
           },
         })
       );
@@ -402,7 +398,6 @@ export function handleIOPub_stream({ msgs, socket, exec_count }) {
           payload: {
             podId,
             content: msgs.content,
-            count: exec_count,
           },
         })
       );
--- a/runtime/src/zmq-utils.ts
+++ b/runtime/src/zmq-utils.ts
@@ -281,12 +281,12 @@ export function handleIOPub_status({ msgs, socket, lang }) {
   );
 }
 
-export function handleIOPub_display_data({ msgs, socket }) {
+export function handleIOPub_display_data({ msgs, socket, exec_count }) {
   console.log("emitting display data ..");
   let [podId, name] = msgs.parent_header.msg_id.split("#");
   let payload = {
     podId,
-    name,
+    //name,
     // content is a dict of
     // {
     //   data: {'text/plain': ..., 'image/png': ...},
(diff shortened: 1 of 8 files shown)


ui/src/components/nodes/Code.tsx
```diff
@@ -278,19 +275,69 @@ export const ResultBlock = memo<any>(function ResultBlock({ id, layout }) {
...
-                {result.text}
-              </Box>
+              {results.map((res, i) => {
+                switch (res.type) {
+                  case "stream_stdout":
+                    return (
+                      <Box
+                        component="pre"
+                        whiteSpace="pre-wrap"
+                        key={i + 1}
```
lihebi, line 295: The key must be unique. There's a console warning.

Screenshot 2023-08-05 at 1 07 50 PM

lihebi:
1. The stderr is not printed.
2. Also, I'd like to keep the visual `<hr/>` to visually separate the stdout/stderr-stream and the return value. 

Reference:

Screenshot 2023-08-05 at 1 12 34 PM

Old behavior:

Screenshot 2023-08-05 at 1 16 23 PM

runtime/src/zmq-utils.ts
```diff
@@ -281,37 +281,35 @@ export function handleIOPub_status({ msgs, socket, lang }) {
...
     //   metadata: {needs_background: 'light'},
     //   transient: ...
     // }
     content: msgs.content,
-    // There's no exe_count in display_data
-    // FIXME I should use execute_reply for count
-    //
-    // count: msgs.content.execution_count,
+    // There's no exec_count in display_data, thus we pass in the session exec_count
+    count: exec_count,
```
lihebi, line 298: Could we pass in 0? This value doesn't seem to be useful. Thus, I don't feel maintaining a session exec_count is necessary. It increases the logic complexity.

lihebi: I think the count in `execute_result` is enough.

ui/src/lib/store/index.tsx
```diff
@@ -27,7 +27,14 @@ export type Pod = {
...
   children: { id: string; type: string }[];
   parent: string;
-  result?: { html?: string; text?: string; count: number; image?: string };
+  result?: {
+    type?: string;
+    html?: string;
+    text?: string;
+    count: number;
+    image?: string;
+  }[];
```
lihebi, line 36: I am not confident about changing this. It is a breaking change, the old values in the DB need migration to work with the new code.

If we really want to change this, we need to supply a DB migration script or procedure/function.

I think the issue you are trying to fix is:
1. the order of stderr
2. being able to display multiple images (at the end, not to be mixed with stdout/stderr streams)

I think you can fix (1) without introducing this schema change. For 2, I'd suggest skipping it for now, it's not crucial.

lihebi: Also, using an array to store streams doesn't sound like a good idea. The stream can come at any granularity, e.g., commonly line-by-line using because `\n` flushes stream in most languages, or users may call `flush()` manually.

I believe "stream" is supposed to be concatenated together upon receiving.

lihebi: In summary, I suggest fixing only issue 1 here, with minimal change so that we don't have to worry about migration. Forget about 2, it's not important.

lihebi: It's better to open another PR to fix issue 1, and leave this PR for future reference.

runtime/src/zmq-utils.ts
senwang86: > I think the count in `execute_result` is enough.

The tricky part is that `execute_result` is not always available, instead, IIUC, `execute_reply` would be the last messages in each cell run. The key is how to update the `count` properly, the logic would be either on the frontend or backend.

ui/src/lib/store/index.tsx
senwang86: > Also, using an array to store streams doesn't sound like a good idea. The stream can come at any granularity, e.g., commonly line-by-line using because `\n` flushes stream in most languages, or users may call `flush()` manually.
> 
> I believe "stream" is supposed to be concatenated together upon receiving.

So the change here is put all the returned messages from kernel in an array, rather than manually separating off each message and concatenating the `text` field. Ipynb kernel would decide how to concatenate each line's execution.

senwang86: > I am not confident about changing this. It is a breaking change, the old values in the DB need migration to work with the new code.
> 
> If we really want to change this, we need to supply a DB migration script or procedure/function.
> 
> I think the issue you are trying to fix is:
> 
> 1. the order of stderr
> 2. being able to display multiple images (at the end, not to be mixed with stdout/stderr streams)
> 
> I think you can fix (1) without introducing this schema change. For 2, I'd suggest skipping it for now, it's not crucial.

IIUC, the JSON format change will not render the `result` filed correctly in the existing repos, in that case, will a re-execution of each pod overwrite the result field?

lihebi: Yes, you are right, the re-execution will fix the result field. The only thing break is the existing result, which might not be that important to do a migration for it, especially at this early release point.

runtime/src/zmq-utils.ts
lihebi: I think we can wait until the final execution results before setting the count. It should be consistent with the behavior of Jupyter?

senwang86: > I think we can wait until the final execution results before setting the count. This should be consistent with the behavior of Jupyter?

Yes, I think a logic in the frontend should work as well.

lihebi: I'm not sure if maintaining `session_exec_count` ourselves is accurate. I'd vote for using `msgs.content.execution_count` because it is always accurate and seems enough for the purpose of showing a count.

lihebi:
> 1. The stderr is not printed.
> 2. Also, I'd like to keep the visual `<hr/>` to visually separate the stdout/stderr-stream and the return value.

What do you say about these two issues? @senwang86  The rest of the code looks good to me.

senwang86:
> > 1. The stderr is not printed.
> > 2. Also, I'd like to keep the visual `<hr/>` to visually separate the stdout/stderr-stream and the return value.
> 
> What do you say about these two issues? @senwang86 The rest of the code looks good to me.

These 2 issues are addressed in the [ab4818c](https://github.com/codepod-io/codepod/pull/436/commits/ab4818c6d256401971dbeec44e69339a47e187f6), can you give it a test?

lihebi:
Just tried, the order is not fixed. Two issues:
1. the order should be `1,2,3`
2. there shouldn't be spaces in between (there are spaces after `3`, and after `2`)

Screenshot 2023-08-07 at 3 54 49 PM

ui/src/components/nodes/Code.tsx
```diff
@@ -278,19 +275,88 @@ export const ResultBlock = memo<any>(function ResultBlock({ id, layout }) {
...
+                          sx={{ fontSize: "0.8em" }}
+                        >
+                          {res.text}
+                        </Box>
+                      );
+                    } else {
+                      return <></>;
+                    }
+                  default:
+                    return <></>;
```
lihebi, line 357: The console warning is still there, caused by these two lines. Adding `key={combineKey}` will fix it.

senwang86:
> Just tried, the order is not fixed. Two issues:
> 
> 1. the order should be `1,2,3`
> 2. there shouldn't be spaces in between (there are spaces after `3`, and after `2`)
> 
> Screenshot 2023-08-07 at 3 54 49 PM

Forget to mention about this, it depends on how Ipynb kernel handles the running result, the result is consistent with Colab. 

Screenshot 2023-08-07 at 4 03 02 PM

lihebi:
I see, SG.

senwang86:
> I see, SG.

Spaces removed. 

Screenshot 2023-08-07 at 4 29 36 PM

lihebi:
Cool, thanks!

PR #440, opened 2023-08-10
feat: use yjs as the source-of-truth

### LINKED ISSUE #331: Yjs as the source of truth
TL;DR Yjs should write the database.

---

The current architecture is centered around the browser (`yjs <-> browser <-> DB`):

- browser (ui/) talk to yjs for peer-syncing
- browser (ui/) talk to DB for data persistence

Since multiple users can have multiple instance of browsers accessing the same repo, this approach has several problems:
- we have some complicated logic to handle when and which user should write the DB.
- there might be consistency issue between Yjs and DB due to bugs

The ideal workflow should be centered around Yjs (`browser <-> Yjs <-> DB`):

- browser talk to yjs for peer-syncing
- Yjs talk to DB for data persistence

lihebi:
This is a major change. Close #331 

## Before
Previous architecture (detailed in #205) is centered around the **browser** (`yjs <-> browser <-> DB`), which is problematic because:
1. we have two (potential inconsistent) sources of data (yjs and DB)
2. there can be multiple users/browsers. It's complex to decide which user/browser should push to DB

## After
Now we are  changing to **Yjs-centered** (`browser <-> Yjs <-> DB`). Yjs is the single source-of-truth.

## Details
This PR removes a lot of syncing logic/functions from front-end, notably those `addPod`, `loadRepo`, `setPodGeo` reducers.

TODOs:
- [ ] Better indicator to show the syncing status (e.g., synced, offline), prompt user unsynced changes when user leave the page.
- [ ] When a user gets disconnected, edits can still be made, but these new edits might be lost. This is a bug. We need to figure out a graceful way to re-sync when Yjs WS re-connects.


lihebi:
This feature is roughly usable. I'm merging this so that people can work on top of this, and avoid merge conflicts, as this PR is too large and touches everything.

Remaining problems: **`state.pods` is now empty**. So the following things break and need fixes in the follow-up PRs:
 * pod.content access, e.g., runtime, parser
 * pod.result access, e.g., display result, stdout/stderr, images
 * export .ipynb and markdown
 * copy & paste

lihebi:
Note: DB schema changed. You need to apply the migration and recreate your docker stack.

ui/src/lib/store/runtimeSlice.tsx
```diff
@@ -22,8 +22,9 @@ function collectSymbolTables(
   let parentId = pod.parent;
   let allSymbolTables: Record<string, string>[] = [];
   // do this for all ancestor scopes.
+  const nodes = Array.from(get().ydoc.getMap<Node>("nodesMap"));
   while (parentId) {
-    let siblings = get().node2children.get(parentId) || [];
+    const siblings = nodes.filter((node) => node.parentNode === parentId);
```
lihebi, line 27: bug: previous siblings are IDs, new siblings are Nodes.

ui/src/lib/store/runtimeSlice.tsx
```diff
@@ -374,12 +377,12 @@ export const createRuntimeSlice: StateCreator<MyState, [], [], RuntimeSlice> = (
     } else if (get().pods[id].type === "SCOPE") {
       // If this pod is a scope, run all pods inside a scope by geographical order.
       // get the pods in the scope
-      let children = get().node2children.get(id);
+      const children = nodes.filter((n) => n.parentNode === id);
```
lihebi, line 380: bug: ditto

lihebi (commented):
A regression bug in two places.

PR #452, opened 2023-08-15
simplify copy-n-paste

lihebi:
New behavior: 
1. select a pod,
2. `ctrl-c` to copy, and
3. `ctrl-v` to paste at the **current mouse position** (the same behavior in [reactflow's example](https://reactflow.dev/docs/examples/interaction/copy-and-paste/)).

TODOs:
- [x] support copying & pasting multiple pods
- [ ] support recursively copying scope pods
- [ ] copy connected edges as well
- [x] copy the content as well, both code and rich pods
- [ ] highlight newly copied pods to indicate that it's not moved into a scope yet
- [ ] support copying results

### DIFF
--- a/ui/src/components/Canvas.tsx
+++ b/ui/src/components/Canvas.tsx
@@ -26,6 +26,7 @@ import ReactFlow, {
   Edge,
   useViewport,
   XYPosition,
+  useStore as useRfStore,
 } from "reactflow";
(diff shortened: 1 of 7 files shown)


lihebi:
Merging, leaving the rest TODOs in follow-up PRs.

PR #458, opened 2023-08-15
[UI] Add Toc in Sidebar

senwang86:
## Summary
- Toc enables the viewport center.
- Will add more follow-up UI tuning in future

## Test
`yarn add @mui/lab @mui/material`

toc

PR #461, opened 2023-08-18
remove CursorNode, fix pod center-on-selection

lihebi:
Two changes:
- #458 introduced a changing behavior: selecting a pod will trigger a centering event. This PR removes that behavior. Centering event is triggered by `state.setCenterSelection(true)`
- `state.CursorNode` (introduced in #431) has been confusing with `state.SelectedPods`. We mentioned that we want to unify the two in https://github.com/codepod-io/codepod/pull/431#issuecomment-1665767266. Therefore, this PR removes `CursorNode` to use only `SelectedPods`.

### DIFF
--- a/ui/src/components/Canvas.tsx
+++ b/ui/src/components/Canvas.tsx
@@ -143,15 +143,23 @@ function isInputDOMNode(event: KeyboardEvent): boolean {
 function useJump() {
   const store = useContext(RepoContext)!;
 
-  const cursorNode = useStore(store, (state) => state.cursorNode);
-  const setCursorNode = useStore(store, (state) => state.setCursorNode);
-
   const setFocusedEditor = useStore(store, (state) => state.setFocusedEditor);
 
   const nodesMap = useStore(store, (state) => state.getNodesMap());
 
+  const reactflow = useReactFlow();
+
+  const selectedPods = useStore(store, (state) => state.selectedPods);
+  const resetSelection = useStore(store, (state) => state.resetSelection);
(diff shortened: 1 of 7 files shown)


lihebi:
Merging, FYI @senwang86

senwang86:
- It used to have an issue by using `SelectedPods` since 1) it doesn't record the order of selection, and 2) it mixes the "single-selection" and "multi-selection". It might be the reason of the following bug,
1. Mouseclick `ScopeB`
2. Left arrow key, which goes to `ScopeA`
3. Shift + Down arrow key, which goes to the first child of `ScopeA`
4. `Esc` should go back to `ScopeA`, __but it went back to `ScopeB`__

jump_bugs

lihebi:
> 4\. `Esc` should go back to `ScopeA`, **but it went back to `ScopeB`**

Why Esc goes to parent? Do you mean shift+up?

lihebi:
There shouldn’t be selection order problems, as there is only one pod being selected in the entire process.

senwang86:
> > 4. `Esc` should go back to `ScopeA`, **but it went back to `ScopeB`**
> 
> Why Esc goes to parent? Do you mean shift+up?

Oh, my bad, forget that we don't use `Esc` during jump.

senwang86:
> There shouldn’t be selection order problems, as there is only one pod being selected in the entire process.

What would happen if you use keyboard to jump after you multi-select pods?

lihebi:
It’s straight forward to determine the target pod by proximity. Then a jump should first reset the selection, then select the target pod.

PR #462, opened 2023-08-18
[UI] Support repo snapshots 

senwang86:
## Summary
- Creating a new database table called `YDocSnapshot` that stores repo snapshots at the request of users.
- The implement features a strict database consistency, once receiving a request, it reads the `yDocBlob` in the `Repo` table and inserts it into `YDocSnapshot`. It avoids conflicts between `Repo.yDocBlob` in the event of network interruption.
- "Rename/restore/delete" operations will be added after discussion.

## Test
- Attach the shell of the `api` docker container 
- Run `npx prisma migrate dev`
- Restart `api` container

add_snapshots

### DIFF
--- a/api/prisma/schema.prisma
+++ b/api/prisma/schema.prisma
@@ -67,6 +67,15 @@ model UserRepoData {
     @@id([userId, repoId])
 }
 
+model YDocSnapshot {
+    id        String   @id
+    createdAt DateTime @default(now())
+    message   String?
+    yDocBlob  Bytes?
+    Repo      Repo?    @relation(fields: [repoId], references: [id])
+    repoId    String?
+}
+
 model Repo {
     id            String         @id
     name          String?
@@ -83,6 +92,8 @@ model Repo {
     UserRepoData  UserRepoData[]
     stargazers    User[]         @relation("STAR")
     yDocBlob      Bytes?
+    yDocSnapshots YDocSnapshot[]
+
 }
 
 enum PodType {
--- a/api/src/resolver_repo.ts
+++ b/api/src/resolver_repo.ts
@@ -518,7 +518,7 @@ async function copyRepo(_, { repoId }, { userId }) {
 /**
  * create yDoc snapshot upon request.
  */
-async function addRepoSnapshot(_, { repoId }, { message }) {
+async function addRepoSnapshot(_, { repoId, message }) {
   const repo = await prisma.repo.findFirst({
     where: { id: repoId },
     include: {
--- a/api/src/resolver_repo.ts
+++ b/api/src/resolver_repo.ts
@@ -515,10 +515,50 @@ async function copyRepo(_, { repoId }, { userId }) {
   return id;
 }
 
+/**
+ * create yDoc snapshot upon request.
+ */
+async function addRepoSnapshot(_, { repoId }, { message }) {
+  const repo = await prisma.repo.findFirst({
+    where: { id: repoId },
+    include: {
+      owner: true,
+      collaborators: true,
+    },
+  });
+  if (!repo) throw Error("Repo not exists.");
+  if (!repo.yDocBlob) throw Error(`yDocBlob on ${repoId} not found`);
+  const snapshot = await prisma.yDocSnapshot.create({
+    data: {
+      id: await nanoid(),
+      yDocBlob: repo.yDocBlob,
+      message: message,
+      repoId: repoId,
+    },
+  });
+  return snapshot.id;
+}
+
+/**
+ * query yDoc snapshot for a repo.
+ */
+async function getRepoSnapshots(_, { repoId }) {
+  const snapshots = await prisma.yDocSnapshot.findMany({
+    where: { repoId: repoId },
+  });
+  if (!snapshots) throw Error(`No snapshot exists for repo ${repoId}.`);
+
+  return snapshots.map((snapshot) => ({
+    ...snapshot,
+    yDocBlob: JSON.stringify(snapshot.yDocBlob),
+  }));
+}
+
 export default {
   Query: {
     repo,
     getDashboardRepos,
+    getRepoSnapshots,
   },
   Mutation: {
     createRepo,
@@ -531,6 +571,7 @@ export default {
     addCollaborator,
     updateVisibility,
     deleteCollaborator,
+    addRepoSnapshot,
     star,
     unstar,
   },
--- a/api/src/typedefs.ts
+++ b/api/src/typedefs.ts
@@ -110,7 +110,7 @@ export const typeDefs = gql`
     repo(id: String!): Repo
     pod(id: ID!): Pod
     getDashboardRepos: [Repo]
-    getRepoSnapshots: [YDocSnapshot]
+    getRepoSnapshots(repoId: String!): [YDocSnapshot]
     activeSessions: [String]
     listAllRuntimes: [RuntimeInfo]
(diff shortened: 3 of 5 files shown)


api/src/typedefs.ts
```diff
@@ -94,6 +94,14 @@ export const typeDefs = gql`
     ttl: Int
   }
 
+  type YDocSnapshot {
+    id: String
+    repoId: String
+    createdAt: String
+    message: String
+    yDocBlob: String
```
lihebi, line 102: This blob doesn’t need to be sent to the front end.

api/src/resolver_repo.ts
```diff
@@ -515,10 +515,50 @@ async function copyRepo(_, { repoId }, { userId }) {
...
+    },
+  });
+  if (!repo) throw Error("Repo not exists.");
+  if (!repo.yDocBlob) throw Error(`yDocBlob on ${repoId} not found`);
+  const snapshot = await prisma.yDocSnapshot.create({
+    data: {
+      id: await nanoid(),
+      yDocBlob: repo.yDocBlob,
+      message: message,
+      repoId: repoId,
```
lihebi, line 536: Use the [connect](https://www.prisma.io/docs/concepts/components/prisma-client/relation-queries#connect-a-single-record) clause to connect the two tables.

api/src/resolver_repo.ts
```diff
@@ -515,10 +515,50 @@ async function copyRepo(_, { repoId }, { userId }) {
...
+  });
+  return snapshot.id;
+}
+
+/**
+ * query yDoc snapshot for a repo.
+ */
+async function getRepoSnapshots(_, { repoId }) {
+  const snapshots = await prisma.yDocSnapshot.findMany({
+    where: { repoId: repoId },
```
lihebi, line 547: {repo: {id: repoId}}

api/prisma/schema.prisma
```diff
@@ -67,6 +67,15 @@ model UserRepoData {
     @@id([userId, repoId])
 }
 
+model YDocSnapshot {
+    id        String   @id
+    createdAt DateTime @default(now())
+    message   String?
+    yDocBlob  Bytes?
+    Repo      Repo?    @relation(fields: [repoId], references: [id])
```
lihebi, line 75: Lower case “repo” field name.

api/src/typedefs.ts
senwang86: > This blob doesn’t need to be sent to the front end.

SG, I was thinking about the "restore" operation, I am not exactly sure if it's better to pass the `yDocBlob` in a single query.

lihebi: Restore should probably happen in the backend yjs server, and front end will retrieve from yjs. I’ll think about this.

api/prisma/schema.prisma
```diff
@@ -67,6 +67,16 @@ model UserRepoData {
     @@id([userId, repoId])
 }
 
+model YDocSnapshot {
+    id        String   @id
+    createdAt DateTime @default(now())
+    message   String?
+    yDocBlob  Bytes?
+    deleted   Boolean  @default(false)
```
lihebi, line 75: I don't want to add `deleted` here since it adds complexity. We could reply on DB backup if we really want to ensure data safety.

senwang86: I am not so familiar with DB backup, any reference for its pros and cons? I am considering a scenario where users might decide to delete a snapshot, but then want to undo the deletion later. Furthermore, if there are 1,000 users performing these operations at different times, can DB backups gracefully address this scenario?

api/prisma/schema.prisma
```diff
@@ -67,6 +67,16 @@ model UserRepoData {
...
+    repo      Repo?    @relation(fields: [repoId], references: [id])
```
lihebi, line 76: yDocBlob and repo/repoId sounds to be required, not optional. I.e., 

```
   yDocBlob Bytes
   repo: Repo
   repoId: String
```

api/prisma/schema.prisma
lihebi: I'd skip this scenario for now. If the user deletes it, it is deleted.

senwang86: > I'd skip this scenario for now. If the user deletes it, it is deleted.

OK, let me revert the change.

api/prisma/schema.prisma
```diff
@@ -71,9 +71,9 @@ model YDocSnapshot {
     id        String   @id
     createdAt DateTime @default(now())
     message   String?
-    yDocBlob  Bytes?
-    repo      Repo?    @relation(fields: [repoId], references: [id])
-    repoId    String?
+    yDocBlob  Bytes
+    repo      Repo    @relation(fields: [repoId], references: [id])
+    repoId    String
 }
```
lihebi, line 77: You need to generate a new DB migration to reflect this change.

senwang86: [bb35923](https://github.com/codepod-io/codepod/pull/462/commits/bb359237593f4a168b3b5151d3888630f9c72efa) addes the prisma migration.

lihebi:
One last comment: merge the two migrations into one? I.e. remove the migrations and re-generate one.

senwang86:
> One last comment: merge the two migrations into one? I.e. remove the migrations and re-generate one.

[d291948](https://github.com/codepod-io/codepod/pull/462/commits/d291948b587332965950f2dd70433b8a3cf63baf) address this.

lihebi:
Awesome, thanks!

lihebi:
Since this PR involves DB schema change, I'd like to ensure that the restoring would work under this design before merging.

I was thinking about how to restore a snapshot. Two methods:
- Option 1 (**not ideal**): Directly replacing the ydoc to a previous state. This isn't ideal, as the peers see a different yDoc with different history, impossible to sync.
- Option 2 (**ideal**): Add a new "restoring" changeset on top of the latest ydoc. This should be the ideal way.

In terms of implementation, I found that Yjs has a ["Y.Snapshot"](https://github.com/yjs/yjs/blob/5ee6992d1fce68993b9a6ea216358d27fc29b294/src/utils/Snapshot.js#L29). So instead of storing a whole YDocBlob into snapshots, we could store a Y.Snapshot binary and use Y.snapshot APIs for restoring. But this snapshot seems to only support method 1, which is not ideal.

To support method 2, we might need to implement a "restoring" changeset, like mentioned [here](https://discuss.yjs.dev/t/snapshots-syncing-and-history/1083/2).

I'll need some time to think about it before we merge this PR and work on restoring.

senwang86:
Realized the fact that the `YDocSnapshot` needs a field of `userId` to separate the snapshots among various collaborators. WDYT, @lihebi ?

lihebi:
I don't think so. The snapshot is associated with the repo. Each repo has only one line of snapshots, regardless of which user triggered it.

senwang86:
> I don't think so. The snapshot is associated with the repo. Each repo has only one line of snapshots, regardless of which user triggered it.

Will the snapshot list show snapshots from all other collaborators?

lihebi:
Yes. There's only one snapshot line for the repo; everyone sees and operates on it. Keep it simple.

lihebi:
There are more things to consider.

I've been thinking about how to make the revision system to:
1. play well with the Git system and
2. to be able to function on scopes

Also, the revision system code wouldn't need to be included in the public-facing desktop app. It would be exclusive to SaaS app, meaning the code would be in another repo.

PR #466, opened 2023-08-20
feat: put runtime behind yjs

lihebi:
Changes:
1. **results are now saved to yDoc** `.getMap("rootMap").get("resultMap")`, so that they sync with peers and are persisted
2. **runtime status is now saved to yDoc** `.getMap("rootMap").get("resultMap")`, so that peers share the exact runtime and share real-time status.
3. the **connection to the runtime kernel is moved behind Yjs**.
    - **Detail**: the front end no longer directly connects to the runtime websocket. Instead, the Yjs server connects to runtime WS, receives and sets results in "resultMap", and sets runtime status in "runtimeMap".
    - **Effect**: if you have some long-running code, **you can now safely close the page. The runtime will continue to run**, and results will continue to be populated (with a TODO caveat[1]). When you reopen the page, you can reload the exact results and runtime status again.
        - [ ] Caveats (TODO): Although the runtime TTL is 12 hours, there's a 30s timeout to keep a YDoc in memory when no one is connected, so for now. These two TTLs shall be unified in the future.
5. Now, you can **create multiple runtimes** and activate either one, see below screenshot:

Screenshot 2023-08-20 at 10 13 53 AM

### DIFF
--- a/api/src/resolver.ts
+++ b/api/src/resolver.ts
@@ -2,6 +2,7 @@ import UserResolver from "./resolver_user";
 import RepoResolver from "./resolver_repo";
 import RuntimeResolver from "./resolver_runtime";
 import ExportResolver from "./resolver_export";
+import YjsResolver from "./resolver_yjs";
 
 export const resolvers = {
   Query: {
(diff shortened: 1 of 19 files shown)


lihebi:
Several TODOs:

- [ ] add text/html output
- [ ] run auto-layout after result changes
       - we had this, but I disabled it in this PR. Shall be added back properly and performantly.

senwang86:
Awesome!

PR #467, opened 2023-08-24
refactor: use pnpm workspace (monorepo)

lihebi:
Several changes:

1. Use monorepo ([ref](https://youtu.be/ngdoUQBvAjo?si=0RTfbkp987zHQwy3), [ref](https://youtu.be/ngdoUQBvAjo?si=xmKrdG3FEqF1eHAz)) via npm/pnpm workspace, so that we can extract code into packages and reuse it across apps (server app, desktop app, across ui/api).

Before:

```
ui/package.json
api/package.json
proxy/package.json
```

After:

```
package.json
pnpm-workspace.yaml
apps/
   ui/package.json
   api/pacakge.json
   proxy/packages.json
packages/
  some-shared-component (TODO)
  some-shared-utils (TODO)
```

2. use `pnpm` instead of `yarn` to manage pacakges. Apart from gaining in the speed and disk space, Yarn actually isn't "right": it treats dependencies as installed packages, thus there has been a lot of "silent" errors. Fixed in this PR.
3. upgrade zustand to 4.4 to fix a bug in package importing https://github.com/pmndrs/zustand/issues/1528


lihebi:
Note for developers:
1. instead of using `yarn` and `yarn add` to install packages (for VSCode linting purpose), use `pnpm install` and `pnpm add ...`. Instead of using `npx prisma ...`, use `pnpx prisma ...`.
2. the docker-compose stack workflow is not changed. But you need to remove your old `ui-node-modules` volumes.

PR #470, opened 2023-08-25
[UI] Fix export import after pr 467

senwang86:
## Summary
- Support html output in pod output display, see screenshot below
- Fix exporting the repo in .ipynb format
- Fix import .ipynb notebook
- To-do:
  - Tune the pod size based on `` size
  - Fix broken `Rich` pod export/import due to usage of yxml
  - Fix "Raw JSON" export function

## Test

Screenshot 2023-08-24 at 9 49 26 PM

### DIFF
--- a/apps/api/src/yjs-runtime.ts
+++ b/apps/api/src/yjs-runtime.ts
@@ -88,6 +88,7 @@ export async function setupRuntimeSocket({
           oldresult.data.push({
             type,
             text: content.data["text/plain"],
+            html: content.data["text/html"],
           });
           resultMap.set(podId, oldresult);
         }
@@ -100,6 +101,7 @@ export async function setupRuntimeSocket({
             type,
             text: content.data["text/plain"],
             image: content.data["image/png"],
+            html: content.data["text/html"],
           });
           resultMap.set(podId, oldresult);
         }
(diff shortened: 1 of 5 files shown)


lihebi:
> To-do:
> * Tune the pod size based on `` size

The image size should be adapted to the pod size, not vice versa. That was the case: if you plot a matplotlib figure on codepod.io, the figure adapts to fit the pod. It broke at some point.


> * Fix "Raw JSON" export function

A side note: as we discussed before, this should be done on the frontend. But more broadly speaking, we need to get a hold on this. This feature is meant to preserve a lossless data of a repo. Since we are using yjs, we need to store yjs content.

lihebi:
The PR looks great, thanks!

senwang86:
> The image size should be adapted to the pod size, not vice versa. That was the case: if you plot a matplotlib figure on codepod.io, the figure adapts to fit the pod. It broke at some point.

Agree, the `matplotlib` figure can be adjusted correctly. While the web images seems not, need some debug later. 

> A side note: as we discussed before, this should be done on the frontend. But more broadly speaking, we need to get a hold on this. This feature is meant to preserve a lossless data of a repo. Since we are using yjs, we need to store yjs content.

Yes, I am aware of this and will discuss the implementation later sometime. Just to put it in the to-do list.

lihebi:
> Agree, the `matplotlib` figure can be adjusted correctly.

Matplotlib figures are not adjusting now. I think it’s broken by #436

PR #471, opened 2023-08-26
[UI][RFC] SelectiveView

senwang86:
## Summary
A prototype of `SelectiveView` via multi-select items in `ToC`
- The Canvas only displays selected pods.
- The `Edit` mode is not affected.

__However__
- The current implementation creates a deep copy of the __selected__ pods from `nodesMap`, thus ALL operations that engage with `nodesMap` need to be properly synced, e.g., moving a pod, creating a new pod, changing the size of a pod, etc. I am not exactly sure if it's the best approach and ask for some comments here. 

## Test
selective_view

### DIFF
--- a/apps/ui/src/components/Sidebar.tsx
+++ b/apps/ui/src/components/Sidebar.tsx
@@ -1294,10 +1294,24 @@ function PodTreeItem({ id, node2children }) {
   );
 }
 
+function dfs(node, node2children, visited) {
+  if (!visited.has(node)) {
+    visited.add(node);
+    const children = node2children.get(node);
+    children.forEach((n) => {
+      dfs(n, node2children, visited);
+    });
+  }
+}
+
 function TableofPods() {
   const store = useContext(RepoContext);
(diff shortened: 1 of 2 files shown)


lihebi:
> * The current implementation renders a deep copy of the **selected** pods from `nodesMap` in `updateView()`

I didn't see "a deep copy". I think you are doing the right thing here, to compute a subset of nodes and `set({nodes})`.

However, I don't quite understand the gif/code:
1. Your gif doesn't show a "partial view" of the canvas. All pods seem to be shown.
2. The sidebar ToC should be "togglable" into three states (selected, hidden, blocked), something like this:

image

senwang86:
> 1. Your gif doesn't show a "partial view" of the canvas. All pods seem to be shown.

It's due to the native multi-selection of tree-view, instead of `checkbox-based` selection, the `selective view` happens when there're more than 1 items selected in ToC.

lihebi:
> the `selective view` happens when there're more than 1 items selected in ToC.

Oh, I see.

PR #474, opened 2023-08-26
[chore] add .dockerignore to avoid copying node_modules

senwang86:
## Summary

It appears that we can avoid copying the `node_modules` folder by using `.dockerignore`, [ref](https://stackoverflow.com/questions/43747776/copy-with-docker-but-with-exclusion).

### DIFF
--- a/apps/api/.dockerignore
+++ b/apps/api/.dockerignore
@@ -1,0 +1,1 @@
+**/node_modules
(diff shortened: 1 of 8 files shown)


lihebi:
If your goal is to improve development stack, this isn’t the place. These dockerfiles are used for production deployment.

PR #475, opened 2023-08-27
[docker-compose] Add setup service to install node_modules and init db

lihebi:
New docker-compose stack behavior:

1. remove xxx-node_modules volumes. We'll install directly on the mounted host dir.
2. add a setup service to run `pnpm install` and initialize DB.

As a result, the new development workflow:

```bash
# step 1 (run once): install node_modules and initialize DB
docker compose up setup

# step 2: fire up the stack
docker compose up -d
```

### DIFF
--- a/compose/dev/.env.example
+++ b/compose/dev/.env.example
@@ -1,0 +1,10 @@
+COMPOSE_PROJECT_NAME=mydev
+
+POSTGRES_USER=myuser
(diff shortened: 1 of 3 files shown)


senwang86:
Seems like `docker-compose down` can't shut down the containers anymore.

lihebi:
You need to use the old `compose.yaml` to shutdown the old stack.

PR #480, opened 2023-08-28
refactor: move prisma to a package

lihebi:
Prisma related code is now moved from `apps/api/` folder to `packages/prisma` folder, so that the prisma client can be shared among different servers.


lihebi:
To use the package in a server app, run `pnpm add @codepod/prisma`, and use like this:

```ts
import prisma from "@codepod/prisma";

prisma.repo.findFirst(...)
```

senwang86:
If you need to perform prisma migration
- Attach the `api` container shell
- run `pnpm dlx prisma migrate dev --schema ../../packages/prisma/prisma/schema.prisma`

lihebi:
> * run `pnpm dlx prisma migrate dev --schema ../../packages/prisma/prisma/schema.prisma`

I would do this instead (looks simpler):

```
cd ../../packages/prisma
pnpm dlx prisma migrate dev
```

PR #488, opened 2023-08-28
[backend] Fix new repo error from PR 485

senwang86:
## Summary
- `Create New Project` in dashboard page throws `resultMap undefined` error and prevents page loading.


lihebi:
Good catch!

PR #489, opened 2023-08-28
[Backend] Reduce resultMap sync frequency

### LINKED ISSUE #483: [Bug] `resultPod. lastExecutedAt` value is missing after execution.
image

where to set PodResults after execution? Let me check.

senwang86:
## Summary
- Reduce yjs sync via `resultMap.set()` method
- Fix  #483; very interesting solution from ChatGPT, while I didn't find its sources
   - _Yjs is a CRDT framework that syncs JavaScript objects across multiple peers. However, it does not natively support all JavaScript types. The Date object is one of the types that Yjs does not natively support. When you try to sync a Date object, Yjs will only sync the reference to the object, not the actual date value. This is why you're seeing issues with syncing the "Date" field._

## Test

Screenshot 2023-08-28 at 2 22 12 PM

### DIFF
--- a/apps/yjs/src/yjs-runtime.ts
+++ b/apps/yjs/src/yjs-runtime.ts
@@ -77,6 +77,7 @@ export async function setupRuntimeSocket({
             type: `${type}_${content.name}`,
             text: content.text,
           });
+          resultMap.set(podId, oldresult);
         }
         break;
       case "execute_result":
@@ -88,6 +89,7 @@ export async function setupRuntimeSocket({
             text: content.data["text/plain"],
             html: content.data["text/html"],
           });
+          resultMap.set(podId, oldresult);
         }
         break;
       case "display_data":
@@ -100,6 +102,7 @@ export async function setupRuntimeSocket({
             image: content.data["image/png"],
             html: content.data["text/html"],
           });
+          resultMap.set(podId, oldresult);
         }
         break;
       case "execute_reply":
--- a/apps/yjs/src/yjs-runtime.ts
+++ b/apps/yjs/src/yjs-runtime.ts
@@ -5,7 +5,6 @@ import { spawnRuntime, killRuntime } from "./runtime";
 
 type PodResult = {
   exec_count?: number;
-  last_exec_end?: boolean;
   data: {
     type: string;
     html?: string;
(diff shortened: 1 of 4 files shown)


apps/yjs/src/yjs-runtime.ts
```diff
@@ -67,6 +67,15 @@ export async function setupRuntimeSocket({
     // FIXME is msg.data a string?
     let { type, payload } = JSON.parse(msg.data as string);
     // console.debug("got message", type, payload);
+    let podId = payload["podId"] || undefined;
+    if (podId) {
+      const oldresult: PodResult = resultMap.get(podId) || { data: [] };
+      if (oldresult.last_exec_end) {
```
lihebi, line 73: I think `last_exec_end` is unnecessary, and I intend to remove it. I was not setting it on purpose.

It was only used to clear the results. The result is now (better) cleared on the frontend right after executing the pod ([code](https://github.com/codepod-io/codepod/blob/579ab3b6c37961c89aafc9405f7063608982bda5/apps/ui/src/lib/store/runtimeSlice.ts#L334)).

senwang86: SG, I missed the [code](https://github.com/codepod-io/codepod/blob/579ab3b6c37961c89aafc9405f7063608982bda5/apps/ui/src/lib/store/runtimeSlice.ts#L334) here. Let me clean up the code.

apps/yjs/src/yjs-runtime.ts
```diff
@@ -78,7 +77,6 @@ export async function setupRuntimeSocket({
             type: `${type}_${content.name}`,
             text: content.text,
           });
-          resultMap.set(podId, oldresult);
```
lihebi, line 81: Is this unnecessary?

lihebi: It is necessary. Otherwise the stream won't be displayed until the very end. Try this:

```py
import time
for i in range(5):
    print(i)
    time.sleep(1)
```

senwang86: I noticed the yjs sync `resultMap` at least 3 times during a single `run`, and intended to reduce the frequency. But from the example above, I agree that yjs should honer the kernel execution behavior even though the sync cost is high. Let me revert the change.

lihebi:
Thanks!

PR #491, opened 2023-08-29
feat: add native-rendered Electron app

lihebi:
There are two ways to make the electron app:

* **[Approach 1, server-based, preferred]**: Host UI at `localhost:3000` and API servers at `localhost:4000`, and use Electron window as a browser. This approach makes it possible to open the app in the browser, so that
   * users can access the browser extensions.
   * lightweight, no need to render another embedded Chromium (the Electron renderer)
   * This approach should also make it easy to re-use the server-app logic.

* **[Approach 2, native-rendered]**: Let host only API server on `localhost:4000`, and render the UI pages & components directly in Electron API. This is the regular way of using Electron.

This PR implements approach 2. It can now render the <Test/> page without a problem. However, approach 1 should be the way to go. ~Therefore, this PR is only for reference.~


lihebi:
> this PR is only for reference.

Actually, this PR is a general platform to build approach 1 as well. I'm merging this.

PR #492, opened 2023-08-29
[UI] Move Runtime status to BottomAppBar

senwang86:
## Summary
A UI refactor which moves the `Sync Status` and `Runtime Status`  from `Sidebar` to a newly created `BottomAppBar`.

Screenshot 2023-08-29 at 4 23 40 PM

## Test

  
bottom_status_bar


apps/ui/src/pages/repo.tsx
```diff
@@ -215,14 +216,19 @@ function RepoWrapper({ children, id }) {
           sx={{
             boxSizing: "border-box",
             width: "100%",
-            height: "100%",
+            height: "96%",
```
lihebi, line 219: This looks fragile.

lihebi:
This UI looks fancy! Instead of "moving", I'd like to keep both (essentially two types of "widgets"). We need a "widget" system to allow users to configure what widgets to show in the sidebar and the bottom-status-bar.

lihebi:
I changed the sidebar just a little bit in #493, resulting in a slight conflict. I've merged that PR since it's pretty big and may cause more conflicts.

lihebi:
A second thought: I probably want to hold on this, because the status items/buttons aren’t fixed, and I’m working on adding locally spawned runtimes to these sections. Working with sidebar seems easier to make changes.

senwang86:
> This UI looks fancy! Instead of "moving", I'd like to keep both (essentially two types of "widgets"). We need a "widget" system to allow users to configure what widgets to show in the sidebar and the bottom-status-bar.

Yeah, I recall this. The only concern is that if we need to maintain 2 sets of code for the same UI system, given that `Sidebar` and `BottomBar` have different space constraint.

senwang86:
> A second thought: I probably want to hold on this, because the status items/buttons aren’t fixed, and I’m working on adding locally spawned runtimes to these sections. Working with sidebar seems easier to make changes.

SG. Just let me know once you finish the local runtime work.

lihebi:
We'll have a widget system that allows users to configure "what" widgets to be displayed at "which" location, be it top or bottom.

senwang86:
Screenshot 2023-08-31 at 6 56 17 PM

apps/ui/src/components/BottomAppBar.tsx
```diff
@@ -0,0 +1,466 @@
...
+    connected: { title: "connected", color: "green" },
+    connecting: { title: "connecting", color: "yellow" },
+    default: { title: "disconnected", color: "red" },
+  };
+  const wsStatus = wsStatusConfig[runtime.wsStatus || "default"];
+
+  return (
+    <ListItem key={runtimeId} disablePadding sx={{ margin: 0 }}>
+      {activeRuntime === runtimeId && (
+        <Tooltip title={wsStatus.title}>
```
lihebi, line 98: Crash when deleting a runtime.

Screenshot 2023-09-01 at 12 54 55 AM

senwang86: Is it tested after merged PR #501 ? it works fine on my end though.

lihebi (commented):
I don't understand how c0b1fd071 creates Sidebar "widget". This commit looks like a few refactorings and a few style changes. It's better if you can create separate small PRs with focused changes.

lihebi:
I used the bottom bar a little, and I feel that this design is a little over-engineered. The first principle of UI design IMO is to be **simple** and **explicit**.

1. Simple: I'm not a fan of replacing text with icons. Texts are simple and easy to understand.
2. Explicit: The bottom bar also hides multiple runtimes into 2nd-level menus, which is not explicit. We may very well need to see all runtime statuses, especially when we introduce dev/prod, training/inference, multiple languages.

Let's use plain text and refine the UI later.

lihebi:
> we need to maintain 2 sets of code for the same UI system

Agree. This is also one of the main concerns that I hesitate to merge. I can feel that this will create headaches for future changes. Let's defer the UI polishing to the very end.

apps/ui/src/components/BottomAppBar.tsx
```diff
@@ -0,0 +1,466 @@
...
+  const store = useContext(RepoContext)!;
+  const runtimeMap =
+    useStore(store, (state) => state.getRuntimeMap()) || new Map();
+  // Observe runtime change
+  const runtimeChanged = useStore(store, (state) => state.runtimeChanged);
+  // A dummy useEffect to indicate that runtimeChanged is used.
+  useEffect(() => {}, [runtimeChanged]);
+  const activeRuntime = useStore(store, (state) => state.activeRuntime);
+  const setActiveRuntime = useStore(store, (state) => state.setActiveRuntime);
+  const runtime = runtimeMap.get(runtimeId) || { wsStatus: "unknown" };
```
lihebi, line 49: [1/3] you set `runtime = {wsStatus: "uknonwn"}` as the fallback

apps/ui/src/components/BottomAppBar.tsx
```diff
@@ -0,0 +1,466 @@
...
+  );
+
+  const runtimeStatusColors = { idle: "green", busy: "yellow", default: "red" };
+  const runtimeStatus = runtime.status ?? "default";
+  const wsStatusConfig = {
+    connected: { title: "connected", color: "green" },
+    connecting: { title: "connecting", color: "yellow" },
+    default: { title: "disconnected", color: "red" },
+  };
+  const wsStatus = wsStatusConfig[runtime.wsStatus || "default"];
```
lihebi, line 93: [2/3] And if that's `unknown`, this wsStatusConfig doesn't have the key, which leads to `wsStatus===undefined`

apps/ui/src/components/BottomAppBar.tsx
lihebi: [3/3] And finally, you access `wsStatus.title` which crashes with "reading properties of undefined" error.

apps/ui/src/components/BottomAppBar.tsx
lihebi: Adding some type annotations can help you catch such bugs early and more comprehensively.

senwang86:
> Thanks @senwang86. My two cents: Can we move the status info to the top bar? The bottom of the screen is usually where the programmer's gaze is at: because code goes top to down. The status bar there sounds a distraction.

Hi @forrestbao , appreciate for the feedback! After some discussion, we agree that the status bar might not offer enough benefit, just as you suggested that it might cause users to be distracted. Thus it will not be merged.

PR #495, opened 2023-08-30
[Backend] Adjust active runtime initialization logic

senwang86:
## Summary
- Find the alive runtime and set as default active runtime
- If `killRuntime()` get called, remove the `runtimeId` from `runtimeMap`

### DIFF
--- a/apps/spawner/src/server.ts
+++ b/apps/spawner/src/server.ts
@@ -129,8 +129,7 @@ export async function startAPIServer({ port }) {
           const doc = getMyYDoc({ repoId, token });
           const rootMap = doc.getMap("rootMap");
           const runtimeMap = rootMap.get("runtimeMap") as Y.Map<RuntimeInfo>;
-          runtimeMap.set(runtimeId, {});
-          // runtimeMap.delete(runtimeId)
+          runtimeMap.delete(runtimeId);
           return true;
         },
 
--- a/apps/ui/src/lib/store/yjsSlice.ts
+++ b/apps/ui/src/lib/store/yjsSlice.ts
@@ -197,10 +197,18 @@ export const createYjsSlice: StateCreator<MyState, [], [], YjsSlice> = (
           });
         }
       );
-      // Set active runtime to the first one.
+      // Set active runtime
       const runtimeMap = get().getRuntimeMap();
       if (runtimeMap.size > 0) {
-        get().setActiveRuntime(Array.from(runtimeMap.keys())[0]);
+        const runtimeList = Array.from(runtimeMap.keys());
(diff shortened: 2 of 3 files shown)


apps/ui/src/lib/store/yjsSlice.ts
```diff
@@ -197,10 +197,18 @@ export const createYjsSlice: StateCreator<MyState, [], [], YjsSlice> = (
...
         }
       );
-      // Set active runtime to the first one.
+      // Set active runtime
       const runtimeMap = get().getRuntimeMap();
       if (runtimeMap.size > 0) {
-        get().setActiveRuntime(Array.from(runtimeMap.keys())[0]);
+        const runtimeList = Array.from(runtimeMap.keys());
+        const activeInstances = runtimeList.filter((id) => {
+          return runtimeMap.get(id)?.status !== undefined;
```
lihebi, line 205: The status must `=== "connected"` to be alive. It could be undefined, "disconnected", or "connecting".

But I'd suggest keeping the original logic (use the first runtime regardless of its status) for now.

There are many remaining logic to add:
1. This assignment only happens at the initial page loading. When there's no runtime, creating a new runtime wouldn't activate it. So we also need to automatically activate the runtime when creating a new runtime (if there's no runtime).
2. We also need to create runtime automatically when creating or opening a repo.

lihebi: I will continue to work on the runtime-related staff. You can focus on the selective view and don't worry too much at the runtime.

senwang86: SG, I noticed that the active runtime continues reset in the UI refactor PR, and realized the root cause here. I will close this PR then.

lihebi: > active runtime continues rese

The runtime isn't stable, but this isn't the root cause, and fixing this alone couldn't make it stable.

apps/spawner/src/server.ts
```diff
@@ -129,8 +129,7 @@ export async function startAPIServer({ port }) {
           const doc = getMyYDoc({ repoId, token });
           const rootMap = doc.getMap("rootMap");
           const runtimeMap = rootMap.get("runtimeMap") as Y.Map<RuntimeInfo>;
-          runtimeMap.set(runtimeId, {});
-          // runtimeMap.delete(runtimeId)
+          runtimeMap.delete(runtimeId);
```
lihebi, line 132: This is a good change, I meant to use `runtimeMap.delete` but forgot to uncomment it. I've actually included this in my local code.

PR #504, opened 2023-09-02
[UI] Unify export functions into one component

senwang86:
## Summany
- Put the export functionality into one components
### Before
Screenshot 2023-09-01 at 6 30 50 PM

### After
Screenshot 2023-09-01 at 6 33 56 PM

### DIFF
--- a/apps/ui/src/components/Sidebar.tsx
+++ b/apps/ui/src/components/Sidebar.tsx
@@ -1155,7 +1155,7 @@ function ExportPanel() {
             backgroundColor: "#1976d2",
           }}
         >
-          Export
+          Export To
         </Paper>
       </Box>
       <CardContent sx={{ margin: "-10px 0px -20px -5px" }}>
@@ -1178,7 +1178,7 @@ function ExportPanel() {
               }}
             >
               <ListItemText>
-                <Typography fontSize={16}>SVG</Typography>
+                <Typography fontSize={16}>Download Image</Typography>
               </ListItemText>
             </ListItemButton>
           </ListItem>
--- a/apps/ui/src/components/Sidebar.tsx
+++ b/apps/ui/src/components/Sidebar.tsx
@@ -43,7 +43,7 @@ import { usePrompt } from "../lib/prompt";
 
 import { RepoContext } from "../lib/store";
 
-import { sortNodes } from "./nodes/utils";
+import { sortNodes, downloadLink, repo2ipynb } from "./nodes/utils";
 
 import {
   FormControlLabel,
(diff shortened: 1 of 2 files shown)


apps/ui/src/components/Sidebar.tsx
```diff
@@ -1092,46 +1102,16 @@ function ExportJupyterNB() {
...
-  if (!store) throw new Error("Missing BearContext.Provider in the tree");
-  const repoName = useStore(store, (state) => state.repoName);
-  const filename = `${repoName?.replaceAll(
-    " ",
-    "-"
-  )}-${repoId}-${new Date().toISOString()}.svg`;
-  const [loading, setLoading] = useState(false);
-
-  const onClick = () => {
+  const exportSVG = () => {
```
lihebi, line 1114: I like the original approach that makes each button a separate React component.

* The original `exportSVG` component is short and easy to understand.
* Now that you embed `exportSVG` into the long `exportJupyter` component, you lose the easy-to-understand SVG logic, and also makes the Jupyter logic even harder to read.

apps/ui/src/components/Sidebar.tsx
```diff
@@ -1148,34 +1128,63 @@ function ExportSVG() {
...
-  );
-}
-
-function ExportButtons() {
-  return (
-    <Stack spacing={1}>
-      <ExportJupyterNB />
-      <ExportSVG />
-    </Stack>
+    <Card elevation={1} sx={{ pl: "0", ml: "-20px", pb: "0", mb: "0" }}>
```
lihebi, line 1141: The `Card` is too complex. Let's use Stack of Buttons like the original one:

```ts
<Stack spacing={1}>
      <ExportJupyterNB />
      <ExportSVG />
</Stack>
```

apps/ui/src/components/Sidebar.tsx
lihebi: If we want to reuse code, we could abstract the common logic into a helper function (e.g., the `downloadLink` function), and use it in both components.

In React, we **compose** components, not **embed** components.

Do this (compose):

```html
<Parent>
  <Child1/>
  <Child2/>
</Parent>
```

Not this (embed):

```html
<Parent>
  <Child1AndChild2/>
</Parent>
```

lihebi: I think the right refactoring is to extract Jupyter export logic into a normal TS function and put it in a `utils.ts` file.

lihebi:
Thanks!

PR #518, opened 2023-09-07
[prototype] move auth out part 1

lihebi:
WIP effort to move user authentication code to only ui-web app.


lihebi:
Implemented in #524

PR #542, opened 2023-09-13
functional refactor & fix: remove state.role

lihebi:
This is one of the efforts to make the component more **"(purely) functional"** by extracting states out. States are global variables that should be avoided when possible.

Removed states:
* `state.role`. Also removed all isGuest checks (replaced by a more functional state.editMode below).
* `state.user`. The user is only used in Awareness and is now passed as function parameter to `connectYjs` action.
* `state.repoLoaded`

Added states:
* `state.editMode: "edit" | "view"`. This could be set on-demand for web-ui when loading user authentication module.

Bug fix:
* In `Code.tsx` and `Rich.tsx`, the Wrap didn't use correct width/height when the role was in "Guest".

### DIFF
--- a/packages/ui/src/components/CanvasContextMenu.tsx
+++ b/packages/ui/src/components/CanvasContextMenu.tsx
@@ -36,11 +36,12 @@ export function CanvasContextMenu(props) {
   const store = useContext(RepoContext);
   if (!store) throw new Error("Missing BearContext.Provider in the tree");
 
-  const isGuest = useStore(store, (state) => state.role === "GUEST");
+  const editMode = useStore(store, (state) => state.editMode);
+  const editing = editMode === "edit";
   return (
     <Box sx={paneMenuStyle(props.x, props.y)}>
       <MenuList className="paneContextMenu">
-        {!isGuest && (
+        {editing && (
           <MenuItem onClick={props.addCode} sx={ItemStyle}>
             <ListItemIcon sx={{ color: "inherit" }}>
               <CodeIcon />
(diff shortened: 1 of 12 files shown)


packages/ui/src/components/CanvasContextMenu.tsx
```diff
@@ -36,35 +36,36 @@ export function CanvasContextMenu(props) {
...
   if (!store) throw new Error("Missing BearContext.Provider in the tree");
 
-  const isGuest = useStore(store, (state) => state.role === "GUEST");
+  const editMode = useStore(store, (state) => state.editMode);
+  const editing = editMode === "edit";
   return (
     <Box sx={paneMenuStyle(props.x, props.y)}>
       <MenuList className="paneContextMenu">
-        {!isGuest && (
+        {editing && (
```
senwang86, line 44: Is it better to control the parent <Box> by `editing`, rather than repeating the `editing` check on each <MenuItem>?

lihebi: There could be some context menu items in "view" mode as well.

packages/ui/src/components/ShareProjDialog.tsx
```diff
@@ -153,22 +153,20 @@ function CollaboratorList({ repoId, collaborators, dispatch, isOwner }) {
...
-                }
-              >
-                <CloseIcon />
-              </IconButton>
-            )
+            <IconButton
+              edge="end"
+              aria-label="delete"
+              onClick={() =>
+                deleteCollaborator({
```
li-xin-yi, line 160: I don't think non-owner user can delete other collaborators. I know it must give an error message from database update operation because of no auth. But in the front-end, we should still disable it.

lihebi: You're right; the current API-auth will throw errors.

I'm trying to remove user/auth-related code from the `packages/ui` so that it can be better re-used across desktop-app and web-app. Things will break. But I consider this as a non-critical break.

Actually, we likely need to add more refined permission control here, e.g., collaborators should be able to add/remove (the behavior of Google Docs).

PR #553, opened 2023-09-22
refactor: rename compose/dev to compose/web

lihebi:
This is a follow-up refactor of #552 to rename compose/dev to compose/web. We used `dev` because we used to have a `prod` folder that was removed in favor of k8s for production.

Now, we have two separated stacks:
* a `compose/web` for a web app with user authentication
* a `compose/desktop` stack without user authentication, and in the future will have local native runtime and file-backend.


lihebi:
Screenshot of the new stacks:

Screenshot 2023-09-21 at 6 25 44 PM

PR #555, opened 2023-09-28
replace pnpm with bun

lihebi:
Pnpm has one major problem: the `.pnpm-store` is not compatible with docker-stack. The problem is due to pnpm's choice to link folders instead of copying. Ref from https://pnpm.io/docker:
> It is impossible to create reflinks or hardlinks between a Docker container and the host filesystem during build time.

This creates the problem that one can either install packages in docker stack, or on the host, but not both. This is pretty inconvenient.

Now bun 1.0 is available, and it seems to work well across the host and docker-stack.

Therefore, this PR uses bun as a package manager to replace pnpm.

TODOs:
- [ ] ~replace node-with-bun.Dockerfile with native oven/bun:1 docker image~. This isn't possible yet due to some compatibility issues with graphql ([ref](https://www.github.com/oven-sh/bun/issues/4947)) & prisma ([ref](https://www.github.com/prisma/prisma/issues/16232#issuecomment-1324821514)). But since I'll probably replace both soon, then we should replace `node-with-bun.Dockerfile` with `oven/bun:1`.
       - [ ] replace primsa with drizzle
       - [ ] replace graphql with tRPC
- [ ] bun seems also a better JS/TS runtime. replace node runtime with bun


lihebi:
Update: the `setup` container is removed. We need to install node packages on the host for IDE linting anyway.

The new developer workflow:

1. install node_modules in your host OS

```
bun install
```

2. migrate DB if this is the first time:

```sh
cd packages/prisma
# e.g., "postgresql://${POSTGRES_USER}:${POSTGRES_PASSWORD}@db:5432/${POSTGRES_DB}?schema=public"
export DATABASE_URL=...
bunx prisma migrate dev
```

3. start docker stack

```
cd compose/web
docker compose up -d
```


TODOS:
- [ ] update developer manual
- [ ] use a hosted DB instead even for development

PR #560, opened 2023-10-04
major refactor: on switching to a local cmd app

lihebi:
The goal is to build a `codepod` cmd tool and use like this:

```shell
> codepod /path/to/repo
# Output:
# codepod has been started, open HTTP://localhost:3000 in your browser
```


lihebi:
A next PR will clean up folders and leave only CMD app related code.

lihebi:
This PR also makes two setup changes:

## 1. remove bun, and use back pnpm (undo #555).

The reason is that bun maintain a **binary lock file**, which makes no sense to me. It makes it impossible to merge PRs.
## 2. do not use docker anymore, for two reasons
1. Docker isn't quite compatible with node packages (not only pnpm, but also npm/yarn for arch-related errors).
2. Also, the source code volume mounting for development often couldn't trigger hot-reloading. As a result, I have to restart the docker stack quite often.

PR #561, opened 2023-10-04
[draft] use nextjs

lihebi:
Why? Because it combines the frontend and backend.
* It's easier for development, as we don't need to fire up multiple apps.
* It's easier for packaging and releasing as a CMD tool (#560).


lihebi:
Update: nextjs doesn't support websocket, so it makes little sense to adopt it.

PR #565, opened 2023-10-06
make codepod (cli) support /path/to/repo; publish codepod to npm

lihebi:
This is a prototype. This PR makes the codepod-cli works with the repoPath cmd-line argument.


lihebi:
Updates:

* building UI package now generates in the api/public folder. This folder is uploaded into npm registry
* yjs-setup is embedded into api/src folder to be able to compile together with api/src code
* adjust express static assets path and UI yjsWsUrl accordingly

End result: now you can install codepod as a npm package:

```shell
# check version
npx @codepod/api --version

# run codepod on a local folder:
npx @codepod/api /path/to/repo
# Output: starting server ..
# html path:  /Users/hebi/.npm/_npx/aa9e62b31be6024c/node_modules/@codepod/api/build/../public
# 🚀 Server ready at http://localhost:4001
```

The files will be saved locally to that folder and persisted when you open codepod again.

TODOs:
- rename package name from `@codepod/api` to  just `codepod`
- better logic to handle exiting

PR #576, opened 2023-11-04
[Prototype] Self-hosted CodeLlama LLM for code autocompletion 

senwang86:
## Summary

This pr provides a solution to use CodeLlama for self-hosted code autocompletion.
- A standalone LLM is required to run separately and expose a RESTful API to access. The most straight-forward way is to use [llama.cpp](https://github.com/ggerganov/llama.cpp), in which you can host the LLM either on a Mac laptop or a GPU-installed machine.
     - Please follow the instructions of [llama.cpp](https://github.com/ggerganov/llama.cpp) to run the LLM locally or in a cloud, also feel free to let me know if a detailed tutorial is needed.
- A Sidebar setting is added to enable/disable the copilot
- tRPC seems not very well support multiple providers in React, [ref](https://github.com/trpc/trpc/discussions/1990), thus I put the Route in `api/src/server`, initially we talked about to create an standalone server service to the copilot-related services.
- There are a couple of follow-up patches to add, e.g., vite bug fix, monitoring the copilot service connection status, similar to the sync statue in the top of Sidebar, adding infilling mode in addition to autocomplete mode. __This PR also aims to collect some early feedback regarding the overall design and architecture.__

## Test
1. First, enable the RESTful API on the llama.cpp, assuming the IP address is `x.x.x.x`, port is `9090`
2. On the local machine, open the terminal, 
     - `cd codepod/api/`
     - `pnpm dev --copilotIP x.x.x.x --copilotPort 9090`
- __Note that, the screenshot below intends to the demonstrate the functionality, the quality of the auto-completion might be low due to the 4-bit quantized llama-7b model__. 

copilot

### DIFF
--- a/api/src/server.ts
+++ b/api/src/server.ts
@@ -10,7 +10,15 @@ import { bindState, writeState } from "./yjs/yjs-blob";
 import cors from "cors";
 import { createSpawnerRouter, router } from "./spawner/trpc";
 
-export async function startServer({ port, repoDir }) {
+export const copilotIpAddress = "127.0.0.1";
+export const copilotPort = 9090;
+
+export async function startServer({
+  port,
+  repoDir,
+  copilotIpAddress = "127.0.0.1",
+  copilotPort = 9090,
+}) {
   console.log("starting server ..");
   const app = express();
   app.use(express.json({ limit: "20mb" }));
@@ -33,6 +41,8 @@ export async function startServer({ port, repoDir }) {
   );
 
   const http_server = http.createServer(app);
+  copilotIpAddress = copilotIpAddress;
+  copilotPort = copilotPort;
 
   // Yjs websocket
   const wss = new WebSocketServer({ noServer: true });
@@ -59,6 +69,8 @@ export async function startServer({ port, repoDir }) {
   });
 
   http_server.listen({ port }, () => {
-    console.log(`🚀 Server ready at http://localhost:${port}`);
+    console.log(
+      `🚀 Server ready at http://localhost:${port}, LLM Copilot is hosted at ${copilotIpAddress}:${copilotPort}`
+    );
   });
 }
--- a/api/src/server.ts
+++ b/api/src/server.ts
@@ -10,9 +10,6 @@ import { bindState, writeState } from "./yjs/yjs-blob";
 import cors from "cors";
 import { createSpawnerRouter, router } from "./spawner/trpc";
 
-export const copilotIpAddress = "127.0.0.1";
-export const copilotPort = 9090;
-
 export async function startServer({
   port,
   repoDir,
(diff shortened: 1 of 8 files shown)


lihebi:
> * tRPC seems not very well support multiple providers in React, [ref](https://github.com/trpc/trpc/discussions/1990),

I also came across multiple providers the other day, and it is well supported: [trpc#3049](https://www.github.com/trpc/trpc/pull/3049). I have implemented multiple providers in https://github.com/codepod-io/codepod-cloud/pull/11. Related code:

https://github.com/codepod-io/codepod-cloud/blob/113f4f7ca3656d6db2296bb32a64dc8ae3ae3342/ui/src/lib/trpc.ts#L9-L16

lihebi:
With that said, it could actually be better and simpler to leave it in `api/` routers, so that the frontend always has one API to talk to. We can let the `api/` forward the request to the actual LLM service internally through tRPC or gRPC.

api/src/server.ts
```diff
@@ -59,6 +68,8 @@ export async function startServer({ port, repoDir }) {
   });
 
   http_server.listen({ port }, () => {
-    console.log(`🚀 Server ready at http://localhost:${port}`);
+    console.log(
+      `🚀 Server ready at http://localhost:${port}, LLM Copilot is hosted at ${copilotIpAddress}:${copilotPort}`
```
lihebi, line 72: I'd revert this print information, because people may choose to run CodePod without copilot server, and this info is misleading. Just be silent should be fine.

ui/src/lib/trpc.ts
```diff
@@ -1,3 +1,21 @@
...
 import type { AppRouter } from "../../../api/src/spawner/trpc";
 export const trpc = createTRPCReact<AppRouter>();
+let remoteUrl;
+
+if (import.meta.env.DEV) {
+  remoteUrl = `localhost:4000`;
+} else {
+  remoteUrl = `${window.location.hostname}:${window.location.port}`;
+}
+export const trpcProxyClient = createTRPCProxyClient<AppRouter>({
```
lihebi: We already have a trpc client in App.tsx. You can access the client in `llamaInlineCompletionProvider` like this:

```ts
// MyMonaco.tsx
function MyMonaco() {
   ...
   const { client } = trpc.useUtils();
   const llamaCompletionProvider = new llamaInlineCompletionProvider(
        id,
        editor,
        client
      );
}
```

lihebi (commented):
Thanks, Sen! It works well. I left some minor comments in the code.

One issue is the automatic completion by `InlineCompletionsProvider` is not very responsiveness. Sometimes it is fired, and sometimes it is not. How about using a shortcut to trigger it manually, and disable the automatic triggering?

lihebi:
Also, there's an uncaught exception in the console for canceled API call. I'd like to catch it and display a canceling message to keep the console clean.

Screenshot 2023-11-07 at 11 50 33 AM

lihebi:
> monitoring the copilot service connection status

This isn't that critical. We can assume that the service is up.



> adding infilling mode in addition to autocomplete mode

This is quite important. It is quite often that we edit code in the middle.

ui/src/lib/trpc.ts
lihebi: A second thought: since the copilot is already a REST API, and we are not going to further customize it or add authentication to this Desktop app, let's directly call the REST API in the frontend.

The trpc is preferred in the cloud app.

senwang86:
After the discussion, we decide to leave this PR as a reference to integrate the self-hosted copilot and address the comments in the codepod-cloud repo.

